PR #23, opened 2021-06-19
Implement Ledger support

### LINKED ISSUE #16: Implement Ledger support
Update: @khorolets is working on it

khorolets:
Closes #16 

Ledger Nano S/X is now supported.

Add Helpers:
* Get publickey from Ledger device 
* Sign a transaction with Ledger
* Sign a transaction with Ledger during simple mode (both send and display are working)

Additionally add missing prompt for signer public key during manual transaction construction.


khorolets:
Also, while I was working on it a newer version of `slip10` was released (included all @frol ) fixes you were relying on, so I updated the `Cargo.toml`

src/commands/utils_command/generate_keypair_subcommand/mod.rs
```diff
@@ -58,7 +39,7 @@ impl CliGenerateKeypair {
                 println!(
                     "Master Seed Phrase: {}\nSeed Phrase HD Path: {}\nImplicit Account ID: {}\nPublic Key: {}\nSECRET KEYPAIR: {}",
                     key_pair_properties.master_seed_phrase,
-                    bip32path_to_string(&key_pair_properties.seed_phrase_hd_path),
+                    crate::common::bip32path_to_string(&key_pair_properties.seed_phrase_hd_path),
```
frol, line 42: HdPath now supports Display and ToString, so we can just drop this helper, and here you just let `key_pair_properties.seed_phrase_hd_path` to be formatted for display.

src/commands/utils_command/generate_keypair_subcommand/mod.rs
```diff
@@ -69,7 +50,7 @@ impl CliGenerateKeypair {
                     "{}",
                     serde_json::to_string_pretty(&serde_json::json!({
                         "master_seed_phrase": key_pair_properties.master_seed_phrase,
-                        "seed_phrase_hd_path": bip32path_to_string(&key_pair_properties.seed_phrase_hd_path),
+                        "seed_phrase_hd_path": crate::common::bip32path_to_string(&key_pair_properties.seed_phrase_hd_path),
```
frol, line 53: We can just use `key_pair_properties.seed_phrase_hd_path.to_string()` here now.

src/commands/utils_command/ledger_publickey_subcommand/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+        );
+        let public_key = match near_ledger::get_public_key(self.seed_phrase_hd_path.clone()).await {
+            Ok(public_key) => public_key,
+            Err(near_ledger_error) => {
+                println!(
+                    "An error occurred while trying to get PublicKey from Ledger device: {:?}",
+                    near_ledger_error
+                );
+                return Ok(());
+            }
```
frol, line 36: Let's return `Err` with the color_eyer Report like here:
 
https://github.com/FroVolod/near-cli/blob/master/src/commands/construct_transaction_command/sign_transaction/sign_with_private_key/mod.rs#L123-L126

src/commands/utils_command/ledger_publickey_subcommand/mod.rs
```diff
@@ -0,0 +1,63 @@
...
+        let implicit_account_id = hex::encode(&public_key);
+
+        match self.format {
+            crate::common::OutputFormat::Plaintext => {
+                println!(
+                    "Seed Phrase HD Path: {}\nImplicit Account ID: {}\nPublic Key: {}",
+                    crate::common::bip32path_to_string(&self.seed_phrase_hd_path),
+                    implicit_account_id,
+                    format!("ed25519:{}", bs58::encode(&public_key).into_string()),
+                );
```
frol, line 48: Let's convert the raw bytes into the near_crypto::PublicKey (I believe we should have TryFrom implementation there. PublicKey type has the Display implementation, so we don't need to do this "magic"

src/commands/utils_command/sign_transaction_with_ledger_subcommand/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+pub struct SignTransactionWithLedger {
+    pub seed_phrase_hd_path: slip10::BIP32Path,
+    pub unsigned_transaction: near_primitives::transaction::Transaction,
+}
+
+impl From<CliSignTransactionWithLedger> for SignTransactionWithLedger {
+    fn from(item: CliSignTransactionWithLedger) -> Self {
+        let seed_phrase_hd_path = match item.seed_phrase_hd_path {
+            Some(hd_path) => hd_path,
+            None => slip10::BIP32Path::from_str("44'/397'/0'/0'/1'").unwrap(),
```
frol, line 24: I believe we should let users enter this custom value in interactive mode. Let's add a method similar to `input_unsigned_transaction` (prepopulate the input with this default value)

src/commands/utils_command/sign_transaction_with_ledger_subcommand/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+                near_crypto::Signature::from_parts(near_crypto::KeyType::ED25519, &signature)
+                    .expect("Signature is not expected to fail on deserialization")
+            }
+            Err(near_ledger_error) => {
+                println!("LEDGER ERROR {:?}", near_ledger_error);
+                color_eyre::Report::msg(format!(
+                    "Transaction is not expected to fail on serialization: {:?}",
+                    near_ledger_error
+                ));
+                return Ok(());
```
frol, line 75: Return it as `Err`, it should report it to the terminal just fine

frol (changes requested):

src/commands/utils_command/ledger_publickey_subcommand/mod.rs
khorolets: That's not a raw bytes. It's `ed25519_dalek::PublicKey`

```
near_crypto::PublicKey::ED25519(
    near_crypto::ED25519PublicKey::from(
        public_key.to_bytes(),
    )
)
```

src/commands/utils_command/sign_transaction_with_ledger_subcommand/mod.rs
khorolets: Disagree, users who will use interactive mode in most cases will use the default hd path. The ones who really need to specify custom hd path will pass it via parameter of command line. WDYT?

src/commands/utils_command/ledger_publickey_subcommand/mod.rs
frol: Yeah, we could benefit from having `From<ed25519_dalek::PublicKey> for ED25519PublicKey`... Meanwhile, your proposed solution is fine 👍

src/commands/utils_command/sign_transaction_with_ledger_subcommand/mod.rs
frol: I considered that as well. Let me share my thinking:

1. It is easy to drop into interactive mode, and you have to make an explicit choice to opt-out of it and ask for `--help`.
2. If you don't see the help, you won't learn about HD path existance, and sometimes you actually need to change it (if you store more than a single key/account)
3. If you need to modify it at any point, you will be forced to craft the whole command, and interactive mode won't be an option for you anymore

Thus, I believe having an option to edit hd path is a good thing. We may just write a good message like "if you are not sure, keep the default" (that is why I proposed to pre-populate the value)

khorolets: Makes sense. I'll update the code asap

src/commands/construct_transaction_command/sign_transaction/sign_with_ledger/mod.rs
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+    pub fn input_seed_phrase_hd_path() -> slip10::BIP32Path {
+        let input: String = Input::new()
+            .with_prompt("Enter seed phrase HD Path (if you not sure leave blank for default)")
+            .default("44'/397'/0'/0'/1'".into())
+            .interact_text()
+            .unwrap();
+        slip10::BIP32Path::from_str(input.as_str()).unwrap()
+    }
```
frol, line 48:
```suggestion
    pub fn input_seed_phrase_hd_path() -> slip10::BIP32Path {
        Input::new()
            .with_prompt("Enter seed phrase HD Path (if you not sure leave blank for default)")
            .with_initial_text("44'/397'/0'/0'/1'".into())
            .interact_text()
            .unwrap()
    }
```

See `with_initial_text` and the avoided use of extra `unwrap()` and just directly read to `BIP32Path`

src/commands/construct_transaction_command/sign_transaction/sign_with_ledger/mod.rs
```diff
@@ -0,0 +1,334 @@
...
+                        &signature,
+                    )
+                    .expect("Signature is not expected to fail on deserialization"),
+                    Err(near_ledger_error) => {
+                        println!("LEDGER ERROR {:?}", near_ledger_error);
+                        color_eyre::Report::msg(format!(
+                            "Transaction is not expected to fail on serialization: {:?}",
+                            near_ledger_error
+                        ));
+                        return Ok(());
```
frol, line 182: Return Err here

frol (approved):

PR #37, opened 2021-07-23
Check account

FroVolod:


src/commands/add_command/access_key/sender/mod.rs
```diff
@@ -47,12 +59,28 @@ impl Sender {
...
+                )? {
+                    Some(_) => break Ok(account_id),
+                    None => println!("This account ID <{}> doesn't exist", account_id),
+                };
+            },
+            None => Ok(Input::new()
+                .with_prompt("What account ID do you need to add a key?")
+                .interact_text()
+                .unwrap()),
+        }
```
frol, line 83: It is less code duplication, and easier to notice the `loop` inside. The downside, though, is that we check if `connection_config` is Some on every iteration, but given that it is not in a tight loop, we can trade this inefficiency for improved readability.

```suggestion
    fn input_sender_account_id(
        connection_config: Option<crate::common::ConnectionConfig>,
    ) -> color_eyre::eyre::Result<String> {
        loop {
            let account_id: String = Input::new()
                    .with_prompt("What account ID do you need to add a key?")
                    .interact_text()
                    .unwrap();
            if let Some(connection_config) = &connection_config {
                if let Some(_) = crate::common::check_account_id(
                    connection_config.clone(),
                    account_id.clone(),
                )? {
                    break Ok(account_id);
                } else {
                    println!("Account <{}> doesn't exist", account_id);
                }
            } else {
                break Ok(account_id);
            }   
        }
```

src/commands/add_command/contract_code/sender/mod.rs
```diff
@@ -25,8 +25,20 @@ impl Sender {
...
-            Some(cli_sender_account_id) => cli_sender_account_id,
-            None => Sender::input_sender_account_id(),
+            Some(cli_sender_account_id) => match &connection_config {
+                Some(network_connection_config) => match crate::common::check_account_id(
+                    network_connection_config.clone(),
+                    cli_sender_account_id.clone(),
+                )? {
+                    Some(_) => cli_sender_account_id,
+                    None => {
+                        println!("This account ID <{}> doesn't exist", cli_sender_account_id);
```
frol, line 35:
```suggestion
                        println!("Account <{}> doesn't exist", cli_sender_account_id);
```

src/commands/transfer_command/transfer_near_tokens_type/mod.rs
```diff
@@ -84,9 +84,37 @@ impl TransferNEARTokensAction {
...
-            None => TransferNEARTokensAction::input_amount(),
+        let amount: crate::common::NearBalance = match &connection_config {
+            Some(network_connection_config) => {
+                let account_balance: u128 = match crate::common::check_account_id(
+                    network_connection_config.clone(),
+                    sender_account_id.clone(),
+                )? {
+                    Some(account_view) => account_view.amount,
+                    None => 0,
+                };
```
frol, line 95: Convert `account_balance` to `NearBalance` once, and then use it instead of passing `u128` around.

```suggestion
                let account_balance: crate::common::NearBalance = match crate::common::check_account_id(
                    network_connection_config.clone(),
                    sender_account_id.clone(),
                )? {
                    Some(account_view) => crate::common::NearBalance::from_yoctonear(account_view.amount),
                    None => crate::common::NearBalance::from_yoctonear(0),
                };
```

src/common.rs
```diff
@@ -318,6 +318,45 @@ impl ConnectionConfig {
...
+                {
+                    result
+                } else {
+                    return Err(color_eyre::Report::msg(format!("Error call result")));
+                };
+            Ok(Some(account_view))
+        }
+        Err(_) => return Ok(None),
+    }
+}
```
frol, line 348: Let's implement a retry logic, so we don't break out of an interactive mode immediately after the first error that could happen due to some network glitch. Retry 5 times with 1 second delay before returning non-Ok(Some) result.

Also, we should notify users that CLI is fetching the data from the server, and so a user should expect to wait a little bit.

frol (changes requested):

PR #40, opened 2021-08-11
construct-transaction/add-contract-code

FroVolod:


src/commands/construct_transaction_command/transaction_actions/add_contract_code_type/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+    pub async fn process(
+        self,
+        prepopulated_unsigned_transaction: near_primitives::transaction::Transaction,
+        network_connection_config: Option<crate::common::ConnectionConfig>,
+    ) -> crate::CliResult {
+        let mut code = Vec::new();
+        std::fs::File::open(&self.file_path.clone())
+            .map_err(|err| color_eyre::Report::msg(format!("Failed to open file: {:?}", err)))?
+            .read_to_end(&mut code)
+            .map_err(|err| color_eyre::Report::msg(format!("Failed to read file: {:?}", err)))?;
```
frol, line 97: You can use `std::fs::read()`: https://doc.rust-lang.org/stable/std/fs/fn.read.html

src/commands/construct_transaction_command/transaction_actions/add_contract_code_type/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+    fn input_file_path() -> std::path::PathBuf {
+        println!();
+        let input_file_path: String = Input::new()
+            .with_prompt("What is a file location of the contract?")
+            .interact_text()
+            .unwrap();
+        let mut path = std::path::PathBuf::new();
+        path.push(input_file_path);
+        println!("path: {:?}", &path);
+        path
```
frol, line 84: You can replace this with:

```suggestion
        input_file_path.into()
```

frol (approved):

src/commands/construct_transaction_command/transaction_actions/add_contract_code_type/mod.rs
FroVolod: fixed

PR #52, opened 2021-08-30
Feature/extension system

volovyk-s:
This code is searching for executables in $PATH if provided subcommand is not found and tries to execute it.
It allows us to build extensions.
Here are some examples of such extensions:
- [Rust + Clap](https://github.com/volovyk-s/near-cli-extension-example)
- [JS + Yargs](https://github.com/volovyk-s/near-cli-extension-example-js) (cool option, but Rust will be the main one)

To try it out:
Rust extension:
```bash
cargo run extension-example rust Cargo.lock
```
This command will search for `rust` word in `Cargo.lock` file

JS Extension:
```bash
cargo run extension-example-js say NEEEAR!
```
This command will print NEEEAR! :)

In the future, we will move helper functions to a separate crate. It will help developers to build good quality extensions and follow our "CLI style" at the same time.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -1254,16 +1254,11 @@ pub fn try_external_subcommand_execution() -> CliResult {
         .map(|dir| dir.join(&subcommand_exe))
         .find(|file| is_executable(file));
 
-    let command = match path {
-        Some(command) => command,
-        None => {
-            return Err(color_eyre::eyre::eyre!(
-                "{} command or {} extension does not exist",
-                subcommand,
-                subcommand_exe
-            ));
-        }
-    };
+    let command = path.ok_or_else(|| color_eyre::eyre::eyre!(
+        "{} command or {} extension does not exist",
+        subcommand,
+        subcommand_exe
+    ))?;
 
     let err = match ProcessBuilder::new(&command).args(&args).exec_replace() {
         Ok(()) => return Ok(()),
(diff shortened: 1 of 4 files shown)


khorolets:
@volovyk-s could you provide example commands how to use it in the PR description, please?

src/common.rs
```diff
@@ -1233,3 +1239,61 @@ mod tests {
         );
     }
 }
+
```
ChaoticTempest, line 1242: nit: It's kind of weird to see these functions being defined after the test module. Usually everything should go before it just for consistency

src/common.rs
```diff
@@ -1233,3 +1239,61 @@ mod tests {
...
+    fs::metadata(path)
+        .map(|metadata| metadata.is_file() && metadata.permissions().mode() & 0o111 != 0)
+        .unwrap_or(false)
+}
+#[cfg(windows)]
+fn is_executable<P: AsRef<Path>>(path: P) -> bool {
+    path.as_ref().is_file()
+}
+
+fn get_path_directories() -> Vec<PathBuf> {
```
ChaoticTempest, line 1293: nit: Normally in rust, we usually only want to use `get` naming if it's the single most obvious way of using it such as `Cell::get`
```suggestion
fn path_directories() -> Vec<PathBuf> {
```

src/main.rs
```diff
@@ -65,7 +64,18 @@ impl Args {
...
+    let cli = match CliArgs::try_parse() {
+        Ok(cli) => cli,
+        Err(error) => {
+            if matches!(
+                error.kind,
+                clap::ErrorKind::UnknownArgument | clap::ErrorKind::InvalidSubcommand
+            ) {
+                return try_external_subcommand_execution();
+            }
+            return Err(color_eyre::eyre::eyre!(error));
```
ChaoticTempest, line 76: This feels a bit funky to have it fail and then go into trying to run an external command, since we'd potentially run into the problem of not giving the right error messages if a user tries to input an invalid command but not a custom command.

Not sure what's the best way of avoiding that situation, but maybe an extra flag/subcommand like `--custom` to tell clap that we're trying to go into our own dynamic subcommand?

ChaoticTempest (changes requested):

src/main.rs
volovyk-s: Now the error is `"unknown command" command or near-cli-"unknown command" extension does not exist`.
Clap allows us to tune it in the future. There is a list of errors that we can handle differently:
https://docs.rs/clap/2.33.0/clap/enum.ErrorKind.html

src/common.rs
```diff
@@ -877,6 +883,62 @@ pub async fn save_access_key_to_keychain(
     Ok(())
 }
 
+pub fn try_external_subcommand_execution() -> CliResult {
+    let mut args: Vec<String> = env::args().skip(1).collect();
+    if args.is_empty() || args[0].is_empty() {
+        return Err(color_eyre::eyre::eyre!("subcommand is not provided"));
+    }
+
+    let subcommand = args.remove(0);
```
frol, line 892:
```suggestion
    let (subcommand, args) = {
        let mut args = env::args().skip(1);
        let subcommand = args.next().ok_or_else(|| color_eyre::eyre::eyre!("subcommand is not provided"))?;
        (subcommand, args.collect())
    };
```

frol (approved):

src/common.rs
volovyk-s: Ok, merged. I have changed the last line to
```rust
 (subcommand, args.collect::<Vec<String>>())
```

ChaoticTempest (approved):

PR #58, opened 2021-09-02
Getting balances to suggest max values in prompt

FroVolod:


FroVolod:
The formula for calculating the max value is not working correctly yet

src/common.rs
```diff
@@ -329,6 +329,45 @@ impl ConnectionConfig {
...
+    let account_items = match check_account_id(connection_config, account_id)? {
+        Some(account_view) => (
+            account_view.amount,
+            account_view.locked,
+            account_view.storage_usage,
+        ),
+        None => (0, 0, 0),
+    };
+    let (amount, locked, storage_usage) = account_items;
+    Ok(NearBalance::from_yoctonear(
```
volovyk-s, line 362: This function does not behave as I expected. It is called `get_max_allowable_transfer_amount` but it returns plain account balance. It means that we can run into a situation like this:
image

Also, I believe that it's better to have a separate function for account balance or account state and then just use it in this function. What I mean (pseudocode):
```rust
func getAccState(acc): AccState {...}

func getMaxTransferAmmount(accState): NearAmount {
    return accState.amount - accState.locked;
}
```

This will allow us to reuse `getAccState` elsewhere.

volovyk-s (changes requested):
In general, I like the idea, but I don't think that we need to show account balance on the screen each time user wants to send tokens. Maybe this part is enough?
```bash
You need to enter a value of no more than 80.615954487870879599999986 NEAR
On your account 80.621874487870879599999986 NEAR. How many NEAR Tokens do you want to transfer? (example: 10NEAR or 0.5near or 10000yoctonear): 1N
```

.github/workflows/release.yml
```diff
@@ -1,8 +1,8 @@
 name: release
 on:
   workflow_dispatch:
-  pull_request:
-    branches: [master]
+  # pull_request:
```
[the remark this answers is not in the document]
FroVolod: Removed a trying to create a release for an unfinished task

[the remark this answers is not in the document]
frol: @volovyk-s CI needs to be further configured to avoid juggling with it on every release. This is what you observe here.

src/common.rs
```diff
@@ -329,6 +329,50 @@ impl ConnectionConfig {
...
+            account_view.locked,
+            account_view.storage_usage,
+        ),
+        None => (0, 0, 0),
+    };
+    let (amount, locked, storage_usage) = account_items;
+    // transaction_costs: u128 = 1_000_000_000_000_000_000_000 - this value is set temporarily
+    // In the future, its value will be calculated by the function: fn tx_cost(...)
+    // https://github.com/near/nearcore/blob/8a377fda0b4ce319385c463f1ae46e4b0b29dcd9/runtime/runtime/src/config.rs#L178-L232
+    let transaction_costs: u128 = 10u128.pow(21);
```
frol, line 365: I suggest renaming it to `pessimistic_transaction_fee`.

src/common.rs
```diff
@@ -329,6 +329,50 @@ impl ConnectionConfig {
...
+        .config_view
+        .runtime_config
+        .storage_amount_per_byte;
+    let account_items = match check_account_id(connection_config, account_id)? {
+        Some(account_view) => (
+            account_view.amount,
+            account_view.locked,
+            account_view.storage_usage,
+        ),
+        None => (0, 0, 0),
```
frol, line 359: It is better to immediately return `Ok(None)` from `get_max_allowable_transfer_amount` instead of encoding the missing data with 0 amount.

src/common.rs
```diff
@@ -329,6 +329,50 @@ impl ConnectionConfig {
     }
 }
 
+pub fn get_max_allowable_transfer_amount(
```
frol, line 332: I suggest renaming to `get_account_transfer_allowance`

src/commands/construct_transaction_command/transaction_actions/transfer_near_tokens_type/mod.rs
```diff
@@ -63,24 +63,38 @@ impl TransferNEARTokensAction {
...
+                    ),
                 }
             }
             None => match item.amount {
                 Some(cli_amount) => cli_amount,
-                None => TransferNEARTokensAction::input_amount(None),
+                None => TransferNEARTokensAction::input_amount(
+                    None,
+                    crate::common::NearBalance::from_yoctonear(0),
+                ),
```
frol, line 97: As a rule of thumb, function name and its arguments should be such so it is easy to guess what the function will do. This function receives None and 0 balance - I have no good guesses based on this piece of information.

P.S. I believe the `input_amount` function does not need the account balance and only needs the transfer allowance as its upper limit.

frol (approved):

src/commands/add_command/stake_proposal/transfer_near_tokens_type/mod.rs
```diff
@@ -180,28 +155,60 @@ impl TransferNEARTokensAction {
 
 impl TransferNEARTokensAction {
```
frol, line 156: I find this whole module screaming for refactoring and renaming (starting from the folder name and ending at the proper use of types and allowance checks)

src/commands/add_command/stake_proposal/transfer_near_tokens_type/mod.rs
```diff
@@ -180,28 +155,60 @@ impl TransferNEARTokensAction {
...
+                        .unwrap();
+                    if let Ok(transfer_amount) = crate::common::TransferAmount::from(
+                        input_amount.clone(),
+                        &account_transfer_allowance,
+                    ) {
+                        break Ok(transfer_amount);
+                    } else {
+                        println!(
+                            "\nWARNING! There is only {} available for stake.",
+                            account_transfer_allowance.transfer_allowance()
```
frol, line 181: there should be a different formula there for staking allowance (`account_liquid_amount + account_locked_amount - transaction_fee`)

frol (changes requested):
P.S. the code duplication is killing me

frol (approved):

src/commands/utils_command/send_signed_transaction/mod.rs
```diff
@@ -64,21 +64,34 @@ impl Transaction {
                     break response;
                 }
                 Err(err) => {
-                    if let Some(serde_json::Value::String(data)) = &err.data {
-                        if data.contains("Timeout") {
-                            println!("Timeout error transaction.\nPlease wait. The next try to send this transaction is happening right now ...");
-                            continue;
+                    match &err.data {
+                        Some(serde_json::Value::String(data)) => {
+                            if data.contains("Timeout") {
```
volovyk-s, line 69: Do we have an error kind here?

src/commands/transfer_command/transfer_near_tokens_type/mod.rs
```diff
@@ -174,28 +149,60 @@ impl TransferNEARTokensAction {
...
+                    ) {
+                        break Ok(transfer_amount);
+                    } else {
+                        println!(
+                            "\nWARNING! There is only {} available for transfer.",
+                            account_transfer_allowance.transfer_allowance()
+                        );
+                        let choose_input = vec![
+                            format!("Yes, I'd like to transfer {}.", input_amount),
+                            "No, I'd like to change change the transfer amount.".to_string(),
```
volovyk-s, line 179: cahnge change

volovyk-s (approved):
I'm still hoping to get an answer to my previous question about UX.

src/commands/transfer_command/transfer_near_tokens_type/mod.rs
FroVolod: fixed

src/commands/utils_command/send_signed_transaction/mod.rs
frol: @volovyk-s We will refactor it once we migrate to https://github.com/near/near-jsonrpc-client-rs

frol:
> I'm still hoping to get an answer to my previous question about UX.

Do you refer to the UX of the available balance? It has been revamped. @FroVolod Please, make a screenshot demonstrating the current flow of reporting the currently available balance + interactions when user enters more than that.

FroVolod:
@frol @volovyk-s here is a demo for transferring tokens with a limit amount: https://asciinema.org/a/mB8MrXDRUqdaSgVzrlMTGXfxq

volovyk-s:
Thanks, @FroVolod ! @frol, I have played with this branch.
I'm referring to this comment: https://github.com/near/near-cli-rs/pull/58#pullrequestreview-745861840
Do we really need to show user balance on each transfer (in terms of security)? 
I suggest checking it only after the amount is entered, or when we have `NOT_ENOUGH_BALANCE` error and only then show the detailed balance. But it's not a strong opinion, just want us to keep it in mind.

frol:
> Do we really need to show user balance on each transfer (in terms of security)?
> I suggest checking it only after the amount is entered, or when we have `NOT_ENOUGH_BALANCE` error and only then show the detailed balance. But it's not a strong opinion, just want us to keep it in mind.

@volovyk-s I would love to learn more about the security concern. Given that signer account id is already on screen, it is trivial to get the balance from the public blockchain anyway. Landing on Wallet main page you are immediately greeted with the amount of USD on your balance, so it is not less secure than that. Printing the balance seems to be user-friendly to me.

PR #64, opened 2021-09-20
refactor: Dedup display functions for view commands

ChaoticTempest:
Deduplicates functions like `display_account_info` and `display_access_key_list`, and puts them into `common.rs`.

This is just one part of the cleanup process, and there will be more after this one. Don't want to snowball all of these into one giant PR, since that wouldn't be good for anyone looking at it, nor would it increase the chance of it being looked at. So let's go with integrating cleanups often

### DIFF
--- a/src/commands/view_command/view_account/block_id/block_id_hash/mod.rs
+++ b/src/commands/view_command/view_account/block_id/block_id_hash/mod.rs
@@ -1,5 +1,9 @@
 use dialoguer::Input;
 
+use crate::common::{display_access_key_list, display_account_info, ConnectionConfig};
+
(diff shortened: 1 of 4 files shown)


volovyk-s (approved):

ChaoticTempest:
is it failing because I need to increment the version number?
```
  with:
    tag_name: 0.1.14
    release_name: 0.1.14
    draft: false
    prerelease: false
  env:
    NEAR_CLI_VERSION: 0.1.14
    GITHUB_TOKEN: ***
Error: Validation Failed: {"resource":"Release","code":"already_exists","field":"tag_name"}
```

frol (approved):

PR #66, opened 2021-09-23
docs: extension system README added

volovyk-s:

Issue #47, opened 2021-08-18
Extract the interactive/cli interface into a standalone crate


frol:
I believe we should be able to reduce the size of the project significantly by abstracting the repetitive implementations with some helpers. This foundation can also be re-used in near CLI extensions (#44)

@volovyk-s I don't have bandwidth to work on this myself, and @FroVolod won't be able to pull it out in a short term since this might involve some derive-macro magic, so someone would need to get up to speed with it, or we might create a bounty for that and invite Rust developers to do it. WDYT?

volovyk-s:
@frol let me discuss it with DevPlatform. I believe that we can work on it starting next quarter.
It may be me, but I'm not sure if I have enough of the Rust experience. Maybe in collaboration with somebody more "rusted", it will be doable.
Anyway, I think it should be done after the new interface is fully merged, to avoid conflicts in code and other complications.
cc @mikedotexe @jim

frol:
It can speed up the development of the new interface significantly, which is why it might be worth exploring it before we move things around. Currently, you need to do quite some manual labor to marry CLI and interactive modes. I will go ahead and informally ask if someone in Russian Rust community is willing to work on it.

mikedotexe:
This is great context. I feel like @ChaoticTempest might be a great person to pair with Serhii on this, but will pass this by Jim.

ChaoticTempest:
@frol we'll be taking this on this sprint and try to implement some kind of proof of concept for now. Will keep this updated as things go

ChaoticTempest:
hey @frol, here's the in progress branch for how this interactive crate will look like: https://github.com/near/near-cli-rs/tree/feature/visual-crate. There's still a good amount of boilerplate since I wanted your opinion on the progress for it before I continue further beyond this.

I haven't removed any code just so most of it still compiles. You'd be able to do the following with this crate at the high level:
```rust
fn main() {
      use near_cli_visual::Interactive;
      let cli = CliArgs::parse().interactive();
}
```
instead of requiring `Args::from` for the interactive bit. Also moved the strum enums to `Cli*` types instead since those would be better associated with the Cli for prompting purposes instead of the non Cli type.

Right now, just to get feedback, this only works so far for the `view_command` branch of this command tree.

The main pattern I've been able to get rid of is the recursive interactive hardcoding via the `Interactive` trait/derive. This will remove the code like the following:
```rust

impl BlockIdHash {
[...]

PR #67, opened 2021-09-23
feat: Validator extension

volovyk-s:
Validator extension works at this point, but I can not say that I'm happy with the result. Here is the path that I suggest we should follow in the future:
- [x] use [near-jsonrpc-client-rs](https://github.com/near/near-jsonrpc-client-rs) instead of `nearcore`
- [ ] use macros from [this](https://github.com/near/near-cli-rs/issues/47) PR (after the merge)
- [ ] create `near-cli-builder` crate inside of this repo, use it in the core project and validators extension (now `validators` extension is a separate project)

In general, 5000+ lines of code for an extension with 2 commands is too much. Even after the creation of macros mentioned above, we will not reach the point when users will be able to build new extensions without pain. I hope that a bunch of macros and libraries mentioned in [this](https://github.com/near/near-cli-rs/pull/66) PR will solve these issues.


frol:
From the first glance, the extension is totally fine from the implementation point of view, though I would spend a few moments thinking through the structure (your README features `near validator list` while the implementation expects `near validator validators`; both variants are a bit odd for different reasons)

> In general, 5000+ lines of code for an extension with 2 commands is too much.

I totally agree. I feel sorry that you had to do it manually. Nobody should struggle over and over again here!

> Even after the creation of macros mentioned above, we will not reach the point when users will be able to build new extensions without pain. I hope that a bunch of macros and libraries mentioned in this PR will solve these issues.

I also have high hopes for that. I hope we can get to the point where we can have so little code that there is nothing to remove from it. Would you simply use clap, your CLI implementation would fit in a few hundred lines. We should definitely challenge ourselves to reach that state with the available automation (derive macros, generics, etc).

frol:
@FroVolod Please, refactor validators extension to interactive-clap and create a separate branch to compare

frol:
@volovyk-s Hey, check out this refactored version using the interactive-clap macro: https://github.com/near/near-cli-rs/tree/validator-extension-interactive-clap/extensions/validator/src

Buckram123:
@frol  Build fails on my machine
```
error: failed to download `near-account-id v0.11.0`

Caused by:
  unable to get packages from source

Caused by:
  failed to parse manifest at `/Users/anymacstore/.cargo/registry/src/github.com-1ecc6299db9ec823/near-account-id-0.11.0/Cargo.toml`

Caused by:
  feature `edition2021` is required

  this Cargo does not support nightly features, but if you
  switch to nightly channel you can add
  `cargo-features = ["edition2021"]` to enable this feature
```
shouldn't `rust-toolchain` be updated to 1.56.0+?

frol:
@Buckram123 This PR should be closed in favor of https://github.com/near/near-cli-rs/tree/validator-extension-interactive-clap/extensions/validator/src

PR #74, opened 2021-10-19
fmt fix

volovyks:


volovyk-s:
@FroVolod let's fix formating, tests are failing.
P.S. Why are we trying to create a release in each PR? And it's failing all the time.

PR #75, opened 2021-11-02
New commands layout

volovyks:
[Direct link to the rendered doc](https://github.com/near/near-cli-rs/blob/new-guide/docs/NEW_NEAR_CLI_INTERFACE.md)


docs/NEW_GUIDE_DRAFT.en.md
```diff
@@ -0,0 +1,57 @@
...
+        - nonce
+    - contract
+        - deploy
+        - dev-deploy
+        - state
+        - code
+        <!-- TODO: what is the best name for it? -->
+        - change-function-call
+        <!-- TODO: what is the best name for it? -->
+        - view-function-call
```
ChaoticTempest, line 20: do we want to remain consistent with near-cli-js? How about something like `modify/mod` and `view`? But docs everywhere mention of `call` so we'd have to change all those references at some point

docs/NEW_GUIDE_DRAFT.en.md
```diff
@@ -0,0 +1,57 @@
...
+    - state
+    - delete
+transfer
+<!-- TODO: should it be manage-account? Should it be here at all? If yes, where is `unstake`? Maybe it should be an extension, like "near-cli-extension-staking".  -->
+stake
+<!-- TODO: feels a bit out of place, but definitely important. Maybe we need to add other view functions and have 1 subcommand for them. -->
+transaction-status
+helpers
+    <!-- TODO: some or all of these helpers can live in extensions, let's discuss it -->
```
ChaoticTempest, line 29: yea I think all of these can live in extensions. But things like `contract-transaction` and `recent-block-hash` should already be installed by default since those seem very essential

docs/NEW_GUIDE_DRAFT.en.md
```diff
@@ -0,0 +1,57 @@
...
+        - state
+        - code
+        <!-- TODO: what is the best name for it? -->
+        - change-function-call
+        <!-- TODO: what is the best name for it? -->
+        - view-function-call
+    - state
+    - delete
+transfer
+<!-- TODO: should it be manage-account? Should it be here at all? If yes, where is `unstake`? Maybe it should be an extension, like "near-cli-extension-staking".  -->
```
ChaoticTempest, line 24: feature creep -- but what if we have a feature where we can have shortcuts? So to be consistent, these would live in `manage-account`, but we also have shortcuts to them?

docs/NEW_GUIDE_DRAFT.en.md
```diff
@@ -0,0 +1,57 @@
...
+    - keys
+        - view
+        - add
+        - delete
+        - nonce
+    - contract
+        - deploy
+        - dev-deploy
+        - state
+        - code
```
volovyk-s, line 16: Should it be code-hash?

docs/NEW_GUIDE_DRAFT.en.md
volovyk-s: maybe `change-call` and `view-call`
or
`call-view`
`call-change`
or
`call-view-function`
`call-change-function`

volovyk-s: I will make it `view-call` and `change-call` for now.

docs/NEW_GUIDE_DRAFT.en.md
```diff
@@ -0,0 +1,68 @@
...
+        - dev-deploy
+        - state
+        - wasm-hash
+        - view-call
+        - change-call
+    - state
+    - delete
+transfer
+<!-- TODO: feels a bit out of place, but definitely important. Maybe we need to add other view functions and have 1 subcommand for them. Or it can be one of the helpers-->
+transaction-status
```
frol, line 26: Maybe we should have some sort of basic explorer command to view transaction status, epoch status, etc.

docs/NEW_GUIDE_DRAFT.en.md
```diff
@@ -0,0 +1,68 @@
...
+    - contract
+        - deploy
+        - dev-deploy
+        - state
+        - wasm-hash
+        - view-call
+        - change-call
+    - state
+    - delete
+transfer
```
frol, line 24: To be honest, I don't like the idea of having `login`, `transfer`, on the top level since they don't match the idea of extensions.

docs/NEW_GUIDE_DRAFT.en.md
```diff
@@ -0,0 +1,68 @@
...
+- validators
+- ..?
+
+Mental model to distinguish extension from the core commands:
+- `NEAR CLI core` is a single binary and contains all of the `core` commands
+- Each extension is a separate binary that can be executed from `NEAR CLI`
+- There is no extensions that are installed `by default`
+- `NEAR CLI` Core commands should be usefull for all groups of users
+- Extensions are used by a particular group or several groups of users
+- Extensions are not composable
```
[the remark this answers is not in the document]
volovyk-s: I mean, that people should not create extensions for extensions.

docs/NEW_GUIDE_DRAFT.en.md
```diff
@@ -0,0 +1,68 @@
...
+
+Open questions:
+- where to add flags like --verbose, --structured, etc.
```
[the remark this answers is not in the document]
volovyk-s: In my head, `--verbose` is for errors, and `--structured` will be used to return structured results, for example, JSON.
Example:
```bash
near account serhii.near balance
```
Result:
```bash
10.1 N
```
```bash
near account serhii.near balance --structured
```
Result:
```bash
{
    "balance": "234567890987654",
    "locked": "7654567654",
    ...
}
```

docs/NEW_GUIDE_DRAFT.en.md
```diff
@@ -0,0 +1,68 @@
...
+- What about prompts like: `Are you sure that you want to delete the last existing key?`. Probably they should be asked after the command is entered.
```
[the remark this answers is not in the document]
volovyk-s: Is there an ability to skip it if we are in the "scripting mode"?

docs/NEW_GUIDE_DRAFT.en.md
volovyk-s: Yes! We were planning explorer as an extension.

docs/NEW_GUIDE_DRAFT.en.md
```diff
@@ -0,0 +1,68 @@
...
+create-account
+    - implicit
+    - subaccount
+<!-- account Id is entered in this step -->
+manage-account
+    - keys
+        - view
+        - add
+        - delete
+        - nonce
```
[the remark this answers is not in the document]
volovyk-s: Agree

docs/NEW_GUIDE_DRAFT.en.md
```diff
@@ -0,0 +1,68 @@
...
+    - local-keys
+        <!-- Needs to be designed (with multikey management) -->
```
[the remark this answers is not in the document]
volovyk-s: At the end of this document, I was trying to create rules to distinguish core functionality from extensions. I believe, that all groups of users will want to manage their local keys (keys that are stored in the `./near-credentials` folder).

docs/NEW_GUIDE_DRAFT.en.md
```diff
@@ -0,0 +1,68 @@
...
+    - contract
+        - deploy
+        - dev-deploy
```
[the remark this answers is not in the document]
volovyk-s: Hm.. it can be a part of development extension. What is your opinion @DiscRiskandBisque ?
And yes, it does not require an account Id, so it's out of place here.

docs/NEW_GUIDE_DRAFT.en.md
```diff
@@ -0,0 +1,68 @@
...
+        - state
+        - wasm-hash
```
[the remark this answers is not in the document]
volovyk-s: Ok, but let's stick to `functions`, not `methods`.
https://docs.near.org/docs/api/rpc/contracts#call-a-contract-function
We are using `function` word in our docs.

docs/NEW_GUIDE_DRAFT.en.md
```diff
@@ -0,0 +1,68 @@
...
+        - view-call
+        - change-call
```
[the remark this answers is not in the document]
volovyk-s: Let's discuss it above.

docs/NEW_GUIDE_DRAFT.en.md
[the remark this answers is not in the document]
frol: Sure, I don't suggest to put it into a separate extension, I just meant that I want it to be extracted from `manage-account` into `manage-keys` or something along those lines (I did not put much thought into the naming)

docs/NEW_GUIDE_DRAFT.en.md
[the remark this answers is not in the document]
frol: The docs use both...
image

Yes, I find it misleading.

@thisisjoshford @amgando What is your opinion on "functions" vs "methods"?

docs/NEW_GUIDE_DRAFT.en.md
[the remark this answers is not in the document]
frol: I see. Well `--structured` is not the best naming since it is not intuitive and also does not communicate explicitly that the output will be in JSON. I don't know any CLI tools that I could get inspiration from for this type of behavior, but I feel it should be defined on the top-level together with `verbose` flag.

I believe we can leave it as an open question, but I would consider to make `--structured` to be a parameter instead of a flag, thus one can potentially choose `human-readable` vs `json` vs `csv` vs ... formats.

docs/NEW_GUIDE_DRAFT.en.md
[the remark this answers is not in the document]
frol: @volovyk-s yes, you have to explicitly add `send` at the end of the command in order to send a transaction without confirmation.

docs/NEW_GUIDE_DRAFT.en.md
frol: I feel I have never had a need to view transaction status from CLI unless I am wearing a smart-contract/blockchain developer hat. If I initiated a transaction from CLI, I will get the report from CLI about the transaction status, so I don't need to view the transaction status manually day-to-day, thus according to our core CLI definition, it should not be part of the near-cli-rs. On the other hand, it could be a small helper, so we may just land it under helpers

docs/NEW_GUIDE_DRAFT.en.md
frol: > where they should be then?

Sorry, I have not put much thought into it yet. Maybe we should put `transfer` under `manage-account` -> `balance` -> `transfer`.

`login` functionality is currently tightly coupled with the NEAR Wallet. There are other ways to get access to your account:

* import account using a seed phrase (we can replicate the Wallet flow of recovering the master key, and then generate and add a new access key that will be stored locally [we should never store the master key pair or seed phrase])
* import account using the key in Ledger (here we have two options: (1) follow the previous approach of adding a new access key [see above] (2) just record that this key is available through Ledger, so near-cli-rs then can hint user to connect Ledger device and follow instructions when one needs to sign a transaction)
* import account using a plaintext private key (I have heard this to be a common feature request for Wallet)
* we might also consider adding flows to create accounts from a linkdrop

Thus, instead of `login` we might want to have a whole top-level command like `import-account`

> And how it contradicts the extensions idea?

top-level commands in my mind should denote some group of actions while `login` and `transfer` are direct actions (this is prone to having more direct actions on the top-level, and then we will be back to dozen options on the top-level)

docs/NEW_GUIDE_DRAFT.en.md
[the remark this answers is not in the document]
volovyk-s: I will move it to Open questions for now. Anyway, we will not implement it at Phase 1.

docs/NEW_GUIDE_DRAFT.en.md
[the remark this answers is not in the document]
volovyk-s: cc @BenKurrek

docs/NEW_GUIDE_DRAFT.en.md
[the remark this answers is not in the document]
volovyk-s: Clarified in this doc.

docs/NEW_GUIDE_DRAFT.en.md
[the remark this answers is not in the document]
BenKurrek: Yea, this should really be more consistent... @bucanero and I are going to go through and restructure the docs so we can fix this. Now the question is should they be called functions or methods. To this, I say that in general, functions are "defined outside of classes while methods are defined inside of and part of classes" as per [this post](https://stackoverflow.com/questions/155609/whats-the-difference-between-a-method-and-a-function#:~:text=A%20very%20general%20definition%20of,of%20and%20part%20of%20classes.)

Read [this](https://www.codecademy.com/articles/fwd-js-methods-functions) also, it's a great article. I think for smart contract programming, it would make more sense to call them methods since they are associated with the smart contract class / structure / object - however you wanna define that.

docs/NEW_GUIDE_DRAFT.en.md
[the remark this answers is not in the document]
volovyk-s: Ok, we can use `--json` and extend it with other formats if needed. Probably `human readable` should be by default. Top-level flag should be the best approach, I believe.
If we will make it a parameter, people will have to choose one of the options every time. I'm not sure is we should complicate things here.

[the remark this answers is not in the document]
volovyk-s: But yes, I will leave it in open questions for now. Closing this discussion for now.

docs/NEW_GUIDE_DRAFT.en.md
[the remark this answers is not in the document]
volovyk-s: In `Rust` it's a bit more tricky: https://doc.rust-lang.org/rust-by-example/fn/methods.html
But in our case contract functions are `methods`.  [Example]( https://github.com/near-examples/rust-status-message/blob/master/src/lib.rs#L23).
In `AS` they are `functions`  because they are not associated with any class. [Example](https://github.com/near-examples/guest-book/blob/master/assembly/main.ts#L13).
But since our main language is `Rust`, let's stick to `methods`.
FYI: we can support more languages in the future, which means that there is no right answer to this question. Let's just be consistent.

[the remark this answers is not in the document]
BenKurrek: yea totally agree.

docs/NEW_GUIDE_DRAFT.en.md
volovyk-s: I gues we can get move `get-recent-block-hash` to `explorer`or `developer` extension too.

docs/NEW_GUIDE_DRAFT.en.md
volovyk-s: Great suggestions about account import options. So `login` becomes `import-account` -> `from-near-wallet`.
I'm adding other options as well.

I agree that `transfer/send` is out of place here.
But we do not have `balance` for `manage-account` -> `balance` -> `transfer` in the current design.
What do you think about  `mange-account` -> `transfer`?

docs/NEW_GUIDE_DRAFT.en.md
```diff
@@ -0,0 +1,81 @@
...
+        - deploy-code
+        - dev-deploy-code
+        - get-code-checksum
+        - get-state
+        - call-view-method
+        - call-change-method
+    - delete
+transfer
+helpers
+    - generate-key
```
volovyk-s, line 39: We can have `off-chain-keys` top-level command that can have these two helpers + local keys management.
Also, we can rename `manage-account` -> `keys` to `manage-account` -> `on-chain-keys`.

frol: Helpers indeed contradict the extensions concept, so I agree with the plan to put these under some top-level command, but I don't have any good ideas at the moment. I don't like the asymmetry of `manage-account -> on-chain-keys` vs just `off-chain-keys`.

docs/NEW_GUIDE_DRAFT.en.md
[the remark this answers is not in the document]
volovyk-s: Moved to dev extension for now.

docs/NEW_GUIDE_DRAFT.en.md
```diff
@@ -0,0 +1,81 @@
...
+    - validators
+    - proposals
+    - ...
+
+Open questions:
+- Where to add flags like `--verbose`, `--structured/json/csv`, etc.
+- We should add `local-keys` management option. Needs to be designed. Can live in `manage-account` -> `local-keys` or at the top level of the `Core NEAR CLI`.
+- Should we have helpers in `Core NEAR CLI`? Is't that contradicts extensions concept?
+
+Other:
```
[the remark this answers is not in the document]
volovyk-s: Ok, added. But they definitely need polishing. Now we have a lot of `manage` and `get`, I'm not sure if that is ok.

docs/NEW_GUIDE_DRAFT.en.md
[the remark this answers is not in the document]
austinabell: > In `Rust` it's a bit more tricky: [doc.rust-lang.org/rust-by-example/fn/methods.html](https://doc.rust-lang.org/rust-by-example/fn/methods.html) But in our case contract functions are `methods`. [Example](https://github.com/near-examples/rust-status-message/blob/master/src/lib.rs?rgh-link-date=2021-11-04T13%3A21%3A48Z#L23). In `AS` they are `functions` because they are not associated with any class. [Example](https://github.com/near-examples/guest-book/blob/master/assembly/main.ts?rgh-link-date=2021-11-04T13%3A21%3A48Z#L13). But since our main language is `Rust`, let's stick to `methods`. FYI: we can support more languages in the future, which means that there is no right answer to this question. Let's just be consistent.

Well, it is methods based on the code that is written in our opinionated framework, but the code generated and what is actually called is actually functions.

The example given here is expanded to:
```
#[cfg(target_arch = "wasm32")]
#[no_mangle]
pub extern "C" fn set_status() {
    near_sdk::env::setup_panic_hook();
   // .. I removed the serde garbage
    let Input { message }: Input = near_sdk::serde_json::from_slice(
        &near_sdk::env::input().expect("Expected input since method has arguments."),
    )
    .expect("Failed to deserialize input from JSON.");
    let mut contract: StatusMessage = near_sdk::env::state_read().unwrap_or_default();
    contract.set_status(message);
    near_sdk::env::state_write(&contract);
}
```

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,114 @@
...
+    - from-near-wallet
+    - from-seed-phrase <seed-phrase>
+    - from-ledger
+    - from-private-key <private-key>
+
+create-account
+    - implicit
+    - subaccount <master-account> <new-subaccount-id>
+
+manage-account <accoundId>
```
frol, line 33: This whole thing is tough, and there is always going to be a tradeoff, so treat my comments as inputs, not the action items.

I feel we diverged from your original proposal of building a resource-oriented design. Should we introduce a top-level `account` command?

* account
  * import
  * create
  * manage

This way it aligns better with `explorer`, `transaction-construction`, etc extensions which are resources, not actions.

There is a problem, however; we can argue that `explorer` can also have `account` resource, and given that almost everything in NEAR is somehow related to accounts, it becomes hard to draw a boundary between those. E.g. is contract manipulation commands are part of account management? (yes and no...)

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,114 @@
...
+        - call-view-method
+        - call-change-method
+    - delete
+
+manage-off-chain-keys
+    - ...
+    - generate-key
+    - get-public-key-from-ledger
+
+manage-connections
```
frol, line 53: Alignment-wise, when I read import-account, create-account, manage-account, manage-keys, and then I see manage-connections I would get confused unless I know that RPCs are involved, and we need to manage those.

I would consider hiding all the CLI configuration commands under a single entrypoint. It seems we can take inspiration from `rustup` tool, which has `rustup self` command to manipulate with its own setup. On another note, we may take inspiration from how `cargo` chooses the compiler version: `cargo build` would use whatever is default, `cargo +beta build` chooses beta version of compiler, so we may have `near +mainnet account manage frol.near keys view`

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,114 @@
...
+```txt
+- validators
+- proposals
+- ...
+```
+
+## Open questions
+- Where to add flags like `--verbose`, `--structured/json/csv`, `--scripting-mode`, etc.
+- How to manage multiple keys? Should we have `default` key? Should they have a name like `mario-game-key-1`?
+- Should we have `extensions` -> `list`/`add`/`delete` commands? We can do it in Phase 2.
```
frol, line 111: I believe we definitely should have a way to manage extensions right from the CLI.

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,114 @@
...
+### `validators` extension
+```txt
+- validators
+- proposals
+- ...
+```
+
+## Open questions
+- Where to add flags like `--verbose`, `--structured/json/csv`, `--scripting-mode`, etc.
+- How to manage multiple keys? Should we have `default` key? Should they have a name like `mario-game-key-1`?
```
frol, line 110: We definitely want to be able to have multiple keys for a single account id. The current near-cli-rs already does that and stores the keys under `~/.near-credentials/<chain-id>/<account-id>/<public-key>.json`

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,114 @@
...
+
+### `validators` extension
+```txt
+- validators
+- proposals
+- ...
+```
+
+## Open questions
+- Where to add flags like `--verbose`, `--structured/json/csv`, `--scripting-mode`, etc.
```
frol, line 109: What is the concern about putting those at the top level?

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,114 @@
...
+- How to manage multiple keys? Should we have `default` key? Should they have a name like `mario-game-key-1`?
+- Should we have `extensions` -> `list`/`add`/`delete` commands? We can do it in Phase 2.
+
+## Other
+- Let's get inspiration for keys and storage data from this project: https://github.com/near-examples/near-account-utils
```
frol, line 114: I don't see any difference from how near-cli-js does stuff. This project just uses `~/.near-credentials` that is created by near-cli-js. What exactly do you want to take as inspiration from this project?

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,114 @@
...
+    - add <connection-name> <network-name> <url1> <url1>
+    - delete <connection-name>
+
+manage-cli-config
+    - set <parameter> <value>
+    - get <parameter>
+```
+
+## Extensions
+Extensions design is a work in progress. They are here mostly to show that we haven't forgotten about particular functionality and that this functionality will not be a part of `Core NEAR CLI`.
```
frol, line 66: Good point, and thanks for the comment!

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,114 @@
...
+```txt
+import-account
+    - from-near-wallet
+    - from-seed-phrase <seed-phrase>
+    - from-ledger
+    - from-private-key <private-key>
+
+create-account
+    - implicit
+    - subaccount <master-account> <new-subaccount-id>
```
frol, line 31: I am not sure how detailed you want to be in this document, but this command will also require the account funding amount in NEAR tokens, the keypair (either specify to generate a new one or ask the user to specify the public key), and at the end, a user would need to sign a transaction (this is in contrast to the implicit account creation flow which does not actually get created on the chain until someone transfers NEAR to it).

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,114 @@
...
+    - get-state
+    - manage-keys
+        - view
+        - add
+        - delete <public-key>
+    - make-transfer <sender> <reciever> <amount>
+    - manage-contract
+        - deploy-code <wasm-path>
+        - get-code-checksum
+        - get-state
```
frol, line 43: How is it different from the `get-state` on the account level?

docs/NEW_GUIDE_DRAFT.en.md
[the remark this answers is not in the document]
DiscRiskandBisque: I agree with frol. `dev-deploy` feels out of place with the other commands. Let's not include it here.

Even in the project extension it could be an option; something like `near project deploy --generateAccounts`.

volovyk-s:
@frol @ChaoticTempest @DiscRiskandBisque 
Here is one more option for the layout of the account commands.
```txt
account
    - import
        - from-near-wallet
        - from-seed-phrase <seed-phrase>
        - from-ledger
        - from-private-key <private-key>
    - create
        - implicit
        - subaccount <master-account> <new-subaccount-id>
    - manage <accoundId>
        - state
            - view
            - transfer <sender> <reciever> <amount>
        - keys
            - view
            - add
            - delete <public-key>
        - contract
            - deploy-code <wasm-path>
            - get-code-checksum
            - get-state
            - call-view-method
            - call-change-method
    - delete
```
In this case, we are avoiding unnecessary `manage` and `get` words in command names. But we are adding one more nested layer.
And I'm still not happy with `near`-> `account` -> `manage` -> `state` -> `view` and `transfer`

docs/NEW_GUIDE_DRAFT.en.md
```diff
@@ -0,0 +1,97 @@
...
+create-account
+    - implicit
+    - subaccount
+manage-account <accoundId>
+    - get-state
+    - manage-keys
+        - view
+        - add
+        - delete
+    - make-transfer
```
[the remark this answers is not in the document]
BenKurrek: would that encompass sending NEAR or would sending NEAR remain in the `near send` command

docs/NEW_GUIDE_DRAFT.en.md
```diff
@@ -0,0 +1,97 @@
+`NEAR CLI` is using `extensions` to saticfy all groups of users.
+    - `Core NEAR CLI` commands should be usefull for all groups of users.
+    - `Core NEAR CLI` is a single binary, there is no extensions that are installed `by default`.
+    - Each extension is a separate binary that can be executed from `NEAR CLI`.
+    - Extensions are used by a particular group or several groups of users.
+    - Extensions are not composable (you should not create extensions for extensions).
```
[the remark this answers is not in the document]
volovyk-s: What I'm trying to say here, is that developers should not create extensions for extensions.
For example, if we will have an extension that will look like this:
```txt
super-extension
    - command-1
    - command-2
```
Then it should not have the ability to be extended with an extension like `near-cli-super-extension-command-3`.
The developer should add it to the `near-cli-super-extension` or create new `near-cli-command3` extension.
Maybe this description is a bit misleading. We can simplify it to `You should not create extensions for extensions`.

docs/NEW_GUIDE_DRAFT.en.md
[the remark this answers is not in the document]
volovyk-s: @frol FT and NFT sound like yet another extension. `transfer-tokens` is ok. I'm more worried about the structure of commands here.
@BenKurrek There is no `near send` in the current design. It was at the top level when this PR was created, but now it somewhere in `near` -> `account` -> `mange` -> `state` -> `transfer-tokens` or `near` -> `manage-account` -> `transfer-tokes`.

docs/NEW_NEAR_CLI_INTERFACE.md
volovyk-s: Can not agree more, but it can be complicated because of the differences in Mac, Windows, and 100500 Linux distributions. Anyway, I think that it's a question for later.

docs/NEW_NEAR_CLI_INTERFACE.md
volovyk-s: Tables and visualization for keys, etc. Check README.

docs/NEW_NEAR_CLI_INTERFACE.md
volovyk-s: Yes, I was thinking about this design.
And yes, there are tradeoffs. Let's not forget, that our CLI will be used by humans. They will come and ask "Where is my `near send` command".

Here is one more option for the layout of the account commands.
```txt
account
    - import
        - from-near-wallet
        - from-seed-phrase <seed-phrase>
        - from-ledger
        - from-private-key <private-key>
    - create
        - implicit
        - subaccount <master-account> <new-subaccount-id>
    - manage <accoundId>
        - state
            - view
            - transfer <sender> <reciever> <amount>
        - keys
            - view
            - add
            - delete <public-key>
        - contract
            - deploy-code <wasm-path>
            - get-code-checksum
            - get-state
            - call-view-method
            - call-change-method
    - delete
```
In this case, we are avoiding unnecessary `manage` and `get` words in command names. But we are adding one more nested layer.
And I'm still not happy with `near`-> `account` -> `manage` -> `state` -> `view` and `transfer`

docs/NEW_NEAR_CLI_INTERFACE.md
volovyk-s: Account state is a representation of this structure in `nearcore`:
```rust
pub struct Account {
    /// The total not locked tokens.
    #[serde(with = "u128_dec_format_compatible")]
    amount: Balance,
    /// The amount locked due to staking.
    #[serde(with = "u128_dec_format_compatible")]
    locked: Balance,
    /// Hash of the code stored in the storage for this account.
    code_hash: CryptoHash,
    /// Storage used by the given account, includes account id, this struct, access keys and other data.
    storage_usage: StorageUsage,
    /// Version of Account in re migrations and similar
    #[serde(default)]
    version: AccountVersion,
}
```

And here, the state is something similar to https://github.com/near-examples/near-account-utils#storage-report. Maybe it should be renamed.

docs/NEW_NEAR_CLI_INTERFACE.md
volovyk-s: `rustup` has only one nested layer:
```bash
SUBCOMMANDS:
    update          Download and install updates to rustup
    uninstall       Uninstall rustup.
    upgrade-data    Upgrade the internal data format.
    help            Prints this message or the help of the given subcommand(s)
```
Our config and connections are more complex, and `near self connections set mainnet` does not sound right to me. `near connections set mainnet` is more clear. But hiding them under one top level command can be a good idea. Maybe `setup` or `config -> connection` + `config -> behaviour`?

IMO, cargo option is even more confusing for non Rust developers.

docs/NEW_NEAR_CLI_INTERFACE.md
frol: Ah, that is indeed not intuitive...

docs/NEW_NEAR_CLI_INTERFACE.md
frol: So far, if we go with resource-oriented top-level commands, `near config connections` looks about right, but I would need to put everything together to see the whole picture before committing to it.

Ok, I don't have a strong opinion about `+network` syntax. Though I still don't see any mention of network selection design in the proposal, so there is nothing to compare with. Should it be something like `--connection mainnet-archival` option at the `near` level?

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,114 @@
...
+    - list-all
+    - add <connection-name> <network-name> <url1> <url1>
+    - delete <connection-name>
+
+manage-cli-config
+    - set <parameter> <value>
+    - get <parameter>
+```
+
+## Extensions
```
frol, line 65: Just remembered two more extensions: `lockup` and `multisig`, and while `lockup` is self-contained, `multisig` is something we might even consider to have integrated into the core CLI. WDYT?

docs/NEW_NEAR_CLI_INTERFACE.md
frol: I think it is not a question of whether we want to ("should we have"), it is just a matter of priority (I believe this should not be the blocker for CLI release).

docs/NEW_NEAR_CLI_INTERFACE.md
frol: I see, so the wording in the document was ambiguous, I thought you wanted to take inspiration about key management / storage layout from that project, but you meant the representation side of it. The current near-cli-rs does a decent job, though I agree that tabular view might work even better:

image

docs/NEW_GUIDE_DRAFT.en.md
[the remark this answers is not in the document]
frol: I tend to agree that from the design perspective FT and NFT should be in dedicated extensions, but it might (yet to be confirmed) be a common use-case; yet, even in that case, I would vote for a dedicated extension and just make sure that installing new extensions is as smooth as possible.

docs/NEW_GUIDE_DRAFT.en.md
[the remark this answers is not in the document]
frol: I see...

> We can simplify it to "You should not create extensions for extensions."

Well, that is still ambiguous as my bash-extension example falls into this group of an extension for extensions, also, it is up to those extensions whether they want to add extra extensibility into their design (e.g. bringing some automation with Lua, JS, whatnot).

I would probably just say that core CLI extensions system only allows to introduce top-level commands through its design.

docs/NEW_NEAR_CLI_INTERFACE.md
ChaoticTempest: for visualization, the https://github.com/mikaelmello/inquire library @volovyk-s and I looked at a while back was pretty promising for inputting data

docs/NEW_NEAR_CLI_INTERFACE.md
volovyk-s: Ok, I will add it to the commands list

docs/NEW_NEAR_CLI_INTERFACE.md
volovyk-s: Ok, it should not be a part of this PR anyway. Cool library @ChaoticTempest , we should give it a try.

docs/NEW_NEAR_CLI_INTERFACE.md
volovyk-s: @frol in this design we are not specifying `connection` in a flag or command. We are setting it in config.
```bash
near config connection set mainnet-1
near account manage serhii.near keys list
```

frol: @volovyk-s that is going to be super annoying if you want to work with localnet / testnet / mainnet day-to-day. Also, setting a config in one terminal will override the behavior of the command in another terminal. This implicitness is exactly what I hate about near-cli-js. I want to have completely reproducible commands.

volovyk-s:
I have moved `import-account` to `local-keys` -> `add-using`, since we are not `importing` the acc, we are actually generating new local key for an existing account, and then adding this key to it. Now it looks like:
```txt
local-keys
    - add-using
        - near-wallet
        - seed-phrase <seed-phrase>
        - ledger
        - private-key <private-key>
    - generate
```

docs/NEW_NEAR_CLI_INTERFACE.md
volovyk-s: I agree that we need to separate `account` into several entities. `contract` is a first candidate for me, since it's not a part of the account, it just belongs to the account.
We can have something similar to `my-account`, probably we can put `actions` there, since they require a key to be executed. It can look like this:
```txt
actions
    - account
        - create
        - delete
    - key
        - add
        - delete
    - transfer
    - stake
    - deploy-contract
    - function-call
```

But with this design, we will put `function-call`s, `keys` and `transfers` under one roof :/

docs/NEW_NEAR_CLI_INTERFACE.md
ChaoticTempest: @frol how about we just have a way to temporarily set the env? We'd still have `near config connection set mainnet-1` as a permanent way to say we want to use mainnet as the default network. But we'd have that other command for the reproducible commands you want. We'd just advise people to use the the temporary config command first, then if they want something more permanent, we'd tell them otherwise

docs/NEW_NEAR_CLI_INTERFACE.md
frol: > But with this design, we will put function-calls, keys and transfers under one roof :/

This is not fun. This is too implementation-driven, and I feel that `actions` is a catch-all name. I wish I could spend more time thinking through it... So far I am not happy with the explored options :disappointed:

docs/NEW_NEAR_CLI_INTERFACE.md
frol: @ChaoticTempest do you mean something line `NEAR_ENV` from near-cli-js? I hate it :smile:

frol:
> I have moved import-account to local-keys -> add-using, since we are not importing the acc, we are actually generating new local key for an existing account, and then adding this key to it.

It is a tough balance between the implementation detail and the actual user expectations / intuition. I like the explicitness of `local-keys` since it will actually educate CLI users and build proper intuition about the process. On the other hand, I feel `add-using` is an odd name, and I feel we miss the distinction between actually generating a new key + storing it on the local FS and adding it to the account AND actually just registering the existing key (let's take "private-key" as an example - you may either just storing the given private key [no transaction to the network is necessary, so it is free of charge], or generate a new key and submit a transaction to add the key to the account signed with this private key [users would need to pay for it and lock more tokens for the storage, but the benefit is that you don't "leak" the "master key"])

mikedotexe (approved):
Seems good to me. Seeing as this is a markdown file, feels like it'll be a great start and iterating is welcome. :)

volovyk-s:
Network selection:
On the first run:
```
What is the default network?
- testnet
- mainnet
Save it to config?
- yes
- no
```

The same can be for `finality`, `default-key`?

frol:
@volovyk-s @ChaoticTempest @FroVolod I have pushed a new commands layout proposal. Here is the [rendered version](https://github.com/near/near-cli-rs/blob/0e298347a4adc0287e969f718d3110ba4226f239/docs/NEW_NEAR_CLI_INTERFACE.md)

It is flat and I hope it is more ergonomic from the end-user perspective.

docs/NEW_NEAR_CLI_INTERFACE.md
volovyk-s: Ok, I will add them to the extensions list for now.

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,191 @@
...
+`NEAR CLI UX principles`
+- All altering actions should have a confirmation step with an option to skip confirmation with an explicit command line parameter (e.g. `send` at the end of the command)
+- All direct children commands of a single parent command should be aligned (either represent an action or a resource, but never a mix of those on the same hierarchy level): `contract` -> `state` (resource) and `contract` -> `deploy` (action) are not aligned, so it should be either `contract` -> `get-state` + `contract` -> `deploy` or `contract` -> `state` -> `view` + `contract` -> `code` -> `deploy`
+- Interactive mode should look like: `command - description`. It will help people to learn the commands.
+
+## `Core NEAR CLI` commands
+
+```
+account
+  - view-account <account-id> network <"mainnet"|"testnet"|...> <now|at-timestamp|at-block-height|at-block-hash>
```
volovyk-s, line 24: Can I assume that `network` and `finality` will be a part of the config? Let's add a list of parameters that we plan to add to the config and place it in this doc.

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,191 @@
...
+ 
+  - create-sub-account <new-account-id> <initial-balance> 
+    (we will treat everything after the first dot in account id as the parent account (transaction signer))
```
volovyk-s, line 27: Great idea!

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,191 @@
...
+    - with-plaintext-public-key "ed25519:..." network <"mainnet"|"testnet"|...>
```
volovyk-s, line 28: So, we will not use the `implicit` word? I like the description, but maybe it should be a part of the command description.

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,191 @@
...
+      - transaction signature options here (see below)
+    - with-generated-keypair network <"mainnet"|"testnet"|...>
```
volovyk-s, line 30: Same as above but key pair will be generated during command execution? It would be cool to merge them, but optional arguments are not supported by the current design.

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,191 @@
...
+    - use-plaintext-public-key "ed25519:..." network <"mainnet"|"testnet"|...>
+      - transaction signature options here (see below)
+    - generate-keypair network <"mainnet"|"testnet"|...>
+      - transaction signature options here (see below)
+
+  - delete-key <account-id> <public-key> network <"mainnet"|"testnet"|...>
+    - transaction signature options here (see below)
+
+  - propose-stake <account-id> <stake-amount-in-NEAR> <validation-node-public-key>
+    (maybe we should extract it into the validators extension)
```
volovyk-s, line 48: I would move it to the validator's extension.  The current design contradicts with `Core NEAR CLI` commands should be useful for all groups of users.` principle described above.

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,191 @@
...
+    - transaction signature options here (see below)
+  
+  - send-nft <nft-contract-account-id> <receiver-account-id> <token-id> network <"mainnet"|"testnet"|...>
+    - transaction signature options here (see below)
+  
+  - view-near-balance network <"mainnet"|"testnet"|...> <now|at-timestamp|at-block-height|at-block-hash>
+  
+  - view-ft-balance <ft-contract-account-id> network <"mainnet"|"testnet"|...> <now|at-timestamp|at-block-height|at-block-hash>
+  
+  - view-nft-list <nft-contract-account-id> network <"mainnet"|"testnet"|...> <now|at-timestamp|at-block-height|at-block-hash>
```
volovyk-s, line 88: Not sure what this command will return.

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,191 @@
...
+
+    - with-init-call <function-name> <function-args> --prepaid-gas <prepaid-gas> --attached-deposit <deposit-amount>
+      - transaction signature options here (see below)
+    - no-init-call
+      - transaction signature options here (see below)
+
+  - download <account-id> to-folder <path-to-download-folder> network <"mainnet"|"testnet"|...> <now|at-timestamp|at-block-height|at-block-hash>
+
+  - view-state <account-id> key-prefix <storage-key-prefix> network <"mainnet"|"testnet"|...> <now|at-timestamp|at-block-height|at-block-hash>
+```
```
volovyk-s, line 71: What about contract hash? Should we add a subcommand for that?

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,191 @@
...
+
+`NEAR CLI UX principles`
+- All altering actions should have a confirmation step with an option to skip confirmation with an explicit command line parameter (e.g. `send` at the end of the command)
+- All direct children commands of a single parent command should be aligned (either represent an action or a resource, but never a mix of those on the same hierarchy level): `contract` -> `state` (resource) and `contract` -> `deploy` (action) are not aligned, so it should be either `contract` -> `get-state` + `contract` -> `deploy` or `contract` -> `state` -> `view` + `contract` -> `code` -> `deploy`
+- Interactive mode should look like: `command - description`. It will help people to learn the commands.
+
+## `Core NEAR CLI` commands
+
+```
+account
```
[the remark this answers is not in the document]
ChaoticTempest: hmm, what about `account -> view`? -- since the context is under `account` anyways, so no need to repeat it. Every other `view-*` would be a more specific form of this.

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,191 @@
...
+  - view-account <account-id> network <"mainnet"|"testnet"|...> <now|at-timestamp|at-block-height|at-block-hash>
+ 
+  - create-sub-account <new-account-id> <initial-balance>
```
ChaoticTempest, line 26: IMO, feel like we can get rid of the dash between sub and account to make `create-subaccount`. That way its more concise to write

docs/NEW_NEAR_CLI_INTERFACE.md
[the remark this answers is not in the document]
frol: > hmm, what about `account -> view`? -- since the context is under `account` anyways, so no need to repeat it. Every other `view-*` would be a more specific form of this.

I don't think we need to sacrifice readability for brevity in near-cli-rs since I believe that we should not expect people to type those commands anyway, and interactive mode proved to be working great.

> I'm not a fan of `view-account` here.

Can you share why? Is it ambiguous or misleading for you?

> Should we rename it to `view-account-info` or something simillar?

We certainly can. We currently name it `account-summary`, so it could also be `view-account-summary`

docs/NEW_NEAR_CLI_INTERFACE.md
frol: I expect to have defaults in the config, but CLI args should always be explicit to be reproducible, so when the argument is not specified, it will just interactively ask with the preselected default, and at the end it will print the full command with those arguments set.

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,191 @@
+# New NEAR CLI interface
+
+`NEAR CLI` is using `extensions` to saticfy all groups of users.
```
BenKurrek, line 3:
```suggestion
`NEAR CLI` is using `extensions` to satisfy all groups of users.
```

docs/NEW_NEAR_CLI_INTERFACE.md
frol: This is not about implicit account. When you create a regular sub-account, you usually want to set the full-access key that will be in control of that account, and so here we have two options: ask user to generate the keypair somehow and specify the public key manually, or let us automate that and generate a new keypair. Well, replying here I realized that I was too low-level here, and I want to take some time thinking on how to make these options user-friendly

docs/NEW_NEAR_CLI_INTERFACE.md
frol: > Same as above but key pair will be generated during command execution?

Yep, that is right.

> It would be cool to merge them, but optional arguments are not supported by the current design.

I don't think we need to merge them. These are different options just like options of how to sign the transaction.

docs/NEW_NEAR_CLI_INTERFACE.md
frol: Yeah, my only concern is that STAKE action then will be the only action we don't represent within the core NEAR CLI. Though, it might still be fine, which is why I left the comment.

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,191 @@
...
+- `Core NEAR CLI` extensions system only allows to introduce top-level commands through its design.
+
+`NEAR CLI` is built for:
+- NEAR `dApp developers`, who build smart-contracts, UIs, and tooling on NEAR.
+- `Tech-savvy` people automating their routines.
+- `Validators`
+
+`NEAR CLI UX principles`
+- All altering actions should have a confirmation step with an option to skip confirmation with an explicit command line parameter (e.g. `send` at the end of the command)
+- All direct children commands of a single parent command should be aligned (either represent an action or a resource, but never a mix of those on the same hierarchy level): `contract` -> `state` (resource) and `contract` -> `deploy` (action) are not aligned, so it should be either `contract` -> `get-state` + `contract` -> `deploy` or `contract` -> `state` -> `view` + `contract` -> `code` -> `deploy`
```
BenKurrek, line 17:
```suggestion
- All direct children commands of a single parent command should be aligned (either represent an action or a resource, but never a mix of those on the same hierarchy level): `contract` -> `state` (resource) and `contract` -> `deploy` (action) are not aligned, so it should be either `contract` -> `state` + `contract` -> `deploy` or `contract` -> `state` -> `view` + `contract` -> `code` -> `deploy`
```

docs/NEW_NEAR_CLI_INTERFACE.md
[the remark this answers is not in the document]
BenKurrek: I agree with @frol here. If this is in the interactive mode, I would be a fan of having more readable commands like `view-account-summary`.

docs/NEW_NEAR_CLI_INTERFACE.md
frol: Do you think it is commonly used and should belong to the core NEAR CLI? I feel it won't hurt to implement, so we can do that, yet, I thought that the `view-account` command could list that information, so we don't need to have a single-purpose low-level command just for this task, and I would extract such commands into a separate contract-development extension.

docs/NEW_NEAR_CLI_INTERFACE.md
frol: I don't have any strong opinion on this, so I am fine renaming it. I will do it on the next iteration

docs/NEW_NEAR_CLI_INTERFACE.md
BenKurrek: I'm a bit confused as to the difference between `with-plaintext-public-key` and `with-generated-keypair` could someone provide some clarification for me?

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,191 @@
...
+
+  - delete-account <account-id> beneficiary <beneficiary-account-id> network <"mainnet"|"testnet"|...>
+    - transaction signature options here (see below)
+
+  - list-keys <account-id> network <"mainnet"|"testnet"|...> <now|at-timestamp|at-block-height|at-block-hash>
+
+  - add-key <account-id>
+    - use-plaintext-public-key "ed25519:..." network <"mainnet"|"testnet"|...>
+      - transaction signature options here (see below)
+    - generate-keypair network <"mainnet"|"testnet"|...>
```
BenKurrek, line 41: Same as my comment above where I'm not too sure the difference between `use-plaintext-public-key` and `generate-keypair`. Also for the `create-subaccount` section, we're using `with-plaintext-public-key` and `with-generated-keypair` but here we're using `use-plaintext-public-key` and just `generate-keypair` so perhaps we can keep it consistent?

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,191 @@
...
+  - propose-stake <account-id> <stake-amount-in-NEAR> <validation-node-public-key>
+    (maybe we should extract it into the validators extension)
+    - transaction signature options here (see below)
+
+  - TODO: "import account" (aka "login") commands
+```
+
+```
+contract
+  - call-view-function <account-id> <function-name> <function-args> network <"mainnet"|"testnet"|...> <now|at-timestamp|at-block-height|at-block-hash>
```
BenKurrek, line 56: I'm not a huge fan of saying `call-view-function` and `call-change-function` because there might be instances where you want to sign a transaction and you're calling a view function. In NAJ and docs, we have it as `viewFunction` and `functionCall`. Perhaps we can mirror this and have `view-function` and `call-function`? In my mind, having `call` makes it seem like you're signing the transaction and having `view` makes it seem like you're not.

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,191 @@
...
+    - transaction signature options here (see below)
+  
+  - deploy <account-id> use-file <path-to-wasm-file>
+
+    - with-init-call <function-name> <function-args> --prepaid-gas <prepaid-gas> --attached-deposit <deposit-amount>
+      - transaction signature options here (see below)
+    - no-init-call
+      - transaction signature options here (see below)
+
+  - download <account-id> to-folder <path-to-download-folder> network <"mainnet"|"testnet"|...> <now|at-timestamp|at-block-height|at-block-hash>
```
BenKurrek, line 68: perhaps we can make this more clear and say `download-wasm` or something because it's not immediately clear what you're downloading

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,191 @@
...
+
+  - view-state <account-id> key-prefix <storage-key-prefix> network <"mainnet"|"testnet"|...> <now|at-timestamp|at-block-height|at-block-hash>
```
BenKurrek, line 70: I wonder if users might get confused as to the difference between `view-state` here and `view-account-info` in the account section

docs/NEW_NEAR_CLI_INTERFACE.md
BenKurrek: It's my understanding that this should return a bunch of `JsonToken` information such as metadata, owner, token ID etc...

To that point, however, should we include from index, limit etc.. into the parameters of this call?

BenKurrek, line 88: Should we include `nft_supply_for_owner` here? This is part of the enumeration standard and would allow you to view how many NFTs you owned.

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,191 @@
...
+  - add-using
+    - near-wallet
+    - seed-phrase <seed-phrase>
+    - ledger
+    - private-key <private-key>
+  - generate
+```
+
+Transaction signature options:
+  * `sign-with-keychain`
```
BenKurrek, line 133: What is sign with `keychain` vs. private key?

docs/NEW_NEAR_CLI_INTERFACE.md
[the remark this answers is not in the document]
volovyk-s: @frol because `view-account` is not descriptive enough. `view-account-summary` is ok.

docs/NEW_NEAR_CLI_INTERFACE.md
volovyk-s: @BenKurrek 
1. `with-plaintext-public-key` - You are creating an account and adding the existing key by providing its public key as a parameter.
2. `with-generated-keypair` - The key will be generated during the execution of this command and saved locally.

BenKurrek: ohhh makes sense - this allows you to have multiple accounts using the same key. Thanks for the clarification!

docs/NEW_NEAR_CLI_INTERFACE.md
volovyk-s: Yeah. If it is a part of `view-account` info - we should not have a separate command.

docs/NEW_NEAR_CLI_INTERFACE.md
volovyk-s: @BenKurrek it breaks the logic of this rule. The `state` is not an action.

docs/NEW_NEAR_CLI_INTERFACE.md
frol: Good point!

How about `view-keyvalue-storage`?

BenKurrek: I like `inspect-storage`! :)

docs/NEW_NEAR_CLI_INTERFACE.md
BenKurrek: Ah ok makes sense. Thanks for the clarification!

docs/NEW_NEAR_CLI_INTERFACE.md
frol: @BenKurrek I would expect CLI would handle pagination internally and query all of the owned NFTs even if it takes multiple function calls.

frol: I expect `view-nft-assets` command to handle that and display the whole list. Do you suggest we have a separate command that will just return a number of owned NFTs?

docs/NEW_NEAR_CLI_INTERFACE.md
frol: Is it any better now?

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,192 @@
...
+
+  - TODO: "import account" (aka "login") commands
+```
+
+```
+contract
+  - call-function
+    - as-read-only <account-id> <function-name> <function-args> network <"mainnet"|"testnet"|...> <now|at-timestamp|at-block-height|at-block-hash>
+    - as-transaction <account-id> <function-name> <function-args> --prepaid-gas <prepaid-gas> --attached-deposit <deposit-amount> network <"mainnet"|"testnet"|...>
+      - transaction signature options here (see below)
```
frol, line 59: @volovyk-s @BenKurrek @miraclx @ChaoticTempest How do you like this layout? I am quite happy about having a single "call-function" which has two clear options, where we will be able to teach the difference in great detail

docs/NEW_NEAR_CLI_INTERFACE.md
BenKurrek: Actually what you said is probably better where you include the number of NFTs owned as well as the assets themselves all in one command

docs/NEW_NEAR_CLI_INTERFACE.md
BenKurrek: I like this a lot - great improvement.

docs/NEW_NEAR_CLI_INTERFACE.md
BenKurrek: I most definitely am a victim of bad NAJ naming haha - I like what you proposed below with one parent command `call-function` and then underneath, you can choose it as read-only or to sign a transaction. Excellent proposition and I'm on board 100% with that.

docs/NEW_NEAR_CLI_INTERFACE.md
BenKurrek: I like this new naming much more. I'm fine with it as is but wanted to know your opinion on the possibility of saying `autogenerate-new-keypair` rather than `autogenerate-keypair`. It's provides a **bit** more context but I'm not sure it's worth it. What are your thoughts @frol

docs/NEW_NEAR_CLI_INTERFACE.md
BenKurrek: Makes sense - what if the user didn't want all that info returned? What if they just wanted to see everything past a starting index of 50 for example?

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,192 @@
...
+
+  - download-wasm <account-id> to-folder <path-to-download-folder> network <"mainnet"|"testnet"|...> <now|at-timestamp|at-block-height|at-block-hash>
+
+  - inspect-storage <account-id> key-prefix <storage-key-prefix> network <"mainnet"|"testnet"|...> <now|at-timestamp|at-block-height|at-block-hash>
+```
+
+```
+tokens <owner-account-id>
+  - send-near <receiver-account-id> <amount-in-NEAR> network <"mainnet"|"testnet"|...>
```
willemneal, line 75: Should there be a `send-yocto` as well?  With `near-units` installed you can do `send-yocto <A> "$(near-units 0.05 N)"`

frol: @willemneal near-cli-rs already supports yN/yNEAR/yoctoNEAR/yoctoN (case-insensitive):

https://github.com/near/near-cli-rs/blob/226bf02bdbc46e2a03c857879d6c2d632e2b27de/src/common.rs#L246-L248

Thus, you can already do `send-near frol.near 1yoctoNEAR`. There is no need to have a separate command for that.

docs/NEW_NEAR_CLI_INTERFACE.md
frol: Good point! I have also extended the command with two options: save the generated key to a keychain or just print it to the terminal.

docs/NEW_NEAR_CLI_INTERFACE.md
volovyk-s: I like it!

docs/NEW_NEAR_CLI_INTERFACE.md
frol: @BenKurrek This is a good point, but I think we can discuss it in more details later. We will need to have more features built into it like filtering and history of NFT operations. If there are <100 NFTs, we don't need to solve the pagination issue and can just fetch all of them, but if there are more than 100 NFTs, it makes no sense to pull all of those anyway, and manually scanning through the pages is not fan either, so we need to identify the use-cases and solve those instead of asking users to paginate though hundreds of NFTs manually.

BenKurrek: Yea for sure - perhaps we can add this to the NFT extension and talk about it when we implement that.

docs/NEW_NEAR_CLI_INTERFACE.md
willemneal: Ah well `<amount-in-NEAR>` made it seem that it only took units of near.  Glad to see there is support.  Though it seems you only support N and yN, but no `mN`.

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -0,0 +1,198 @@
...
+  - delete-key <account-id> <public-key> network <"mainnet"|"testnet"|...>
+    - transaction signature options here (see below)
+
+  - TODO: "import account" (aka "login") commands
+```
+
+```
+contract
+  - call-function
+    - as-read-only <account-id> <function-name> <function-args> network <"mainnet"|"testnet"|...> <now|at-timestamp|at-block-height|at-block-hash>
```
willemneal, line 63: Why not just `view`?  Even knowing that as-read-only means `view` I found it confusing and is less intuitive since we use `view` everywhere else.

```
near-cli contract call-function as-read-only ...
```

Is also a bit long.  I know you're love for being explicit, but having some aliases here would be nice.

```
near contract view
``` 

As a short cut would be very handy.

docs/NEW_NEAR_CLI_INTERFACE.md
frol: We could add more prefixes in the future when there is evidence that it is necessary. I have not seen use-cases when someone wants to attach or send mN amounts.

willemneal: It comes in handy when you need to provide a small amount to cover storage.  But I agree it's a rare instance.

docs/NEW_NEAR_CLI_INTERFACE.md
BenKurrek: To me just saying `view` might lead some people to think that they're viewing the contract code rather than it being viewing the result of a read-only function.

willemneal: If you execute a view call and try to invoke a host function like creating a transaction, you will get an error that says it was banned in a view call.  So my point is that we use `view` in our documentation and baked into the runtime, so using it, especially in this context helps people to learn it.

frol: @willemneal If you finish up the whole command, you will see that it does not make any sense to save a few characters:

```
near-cli contract view zavodil.poolv1.near get_owner_id '{}' network mainnet now
near-cli contract call-function as-read-only zavodil.poolv1.near get_owner_id '{}' network mainnet now
```

I don't expect anyone to type it out manually as we have interactive mode that guides users through it, and prints out the final command if you want to re-use it later.

`contract view` makes sense only to victims of near-cli-js naming. Anyone else will be confused and guess that this command will show you the contract, not calling a function. Yet, the good part of this new approach is that you will be able to implement your own extension and call it as `near-cli view ...` just fine, if you feel you want to have a shortcut.

docs/NEW_NEAR_CLI_INTERFACE.md
frol: BTW, if you will ever feel limited by the supported prefixes and want to use `near-units`, you can still use

```
send-near <A> "$(near-units 5 mN) yoctoNEAR"
```

frol (approved):

BenKurrek (approved):

willemneal:
My only nit left and it's fine if everyone else thinks otherwise, but `view` is used throughout the ecosystem so not including it will confuse people.  So I propose the following change.  

`near-cli contract call-function as-read-only `--> `near-cli contract call-function as-view`

I'll go ahead and approve, but just wanted to give one more argument for not removing view.

willemneal (approved):

BenKurrek:
> My only nit left and it's fine if everyone else thinks otherwise, but `view` is used throughout the ecosystem so not including it will confuse people. So I propose the following change.
> 
> `near-cli contract call-function as-read-only `--> `near-cli contract call-function as-view`
> 
> I'll go ahead and approve, but just wanted to give one more argument for not removing view.

I'm fine with this as well - I don't have a strong opinion as to which I prefer. I'm fine with both.

docs/NEW_NEAR_CLI_INTERFACE.md
kalloc: Hi, very nice PR. But i have a question.
Why not to use  `view` and `call` as argument for contract calling, I think not all people understand how read-only call works and `call-function as-read-only` may misunderstood people

frol:
So far I feel ok with going `contract <contract-account-id> call-function as-view`, but I want to take a bit of a time and also wait for the feedback from other folks.

I forgot to add commands for offline transaction construction options (in order to construct a transaction you need to specify recent block hash and access key nonce), so I will finish that up before merging this PR.

docs/NEW_NEAR_CLI_INTERFACE.md
```diff
@@ -42,15 +42,18 @@ account
...
   - add-key <account-id>
-    - autogenerate-new-keypair network <"mainnet"|"testnet"|...>
-      - save-to-keychain network <"mainnet"|"testnet"|...>
+    - grant-full-access
+      - autogenerate-new-keypair
+        - save-to-keychain network <"mainnet"|"testnet"|...>
+          - transaction signature options here (see below)
+        - print-to-terminal network <"mainnet"|"testnet"|...>
+          - transaction signature options here (see below)
+      - use-manually-provided-seed-phrase "twelve words goes here" network <"mainnet"|"testnet"|...>
```
willemneal, line 51: This is an issue I filled with near-cli.  It's bad security practice to use a seed phrase as a CLI arg. So perhaps this should be interactive like a password.

frol: It is both interactive and non-interactive, and again, I don’t expect users to type the command out manually most of the time, so the parameters are there only for reproducibility and integration within scripts, which I don’t see a reason to limit to use params

frol:
FYI, This new design is already mostly implemented in a separate repo: https://github.com/FroVolod/near-cli-rs-instance/tree/without-generics

frol:
@volovyks @ChaoticTempest Thanks for all your valuable input along the way!

Check out the [0.2.0-pre.1 release](https://github.com/near/near-cli-rs/releases/tag/0.2.0-pre.1)!

PR #77, opened 2021-11-18
Interactive Clap parsing

ChaoticTempest:
Roughly working `#[derive(Eclap)]`.

naming isn't final, but here's the dictionary:
- `eclap` => "extended clap"
- `iparse` => "interactive parse"

This will let us create interactive clap macros that will fill in cli commands with either `prompt_msg` or `prompt_fn` if they're not filled in when we go to enter into the command line: `near ...`
```rust
#[derive(Debug, Eclap)]
struct TopLevel {
    #[eclap(prompt_msg = "To bar or not to?")]
    bar: bool,
    
    #[eclap(single, subcommand)]
    qux: MySubcommand,
}

#[derive(Debug, Eclap)]
struct MySubcommand {
     #[eclap(prompt_msg = "To qux or not to?")]
     qux: bool
}
```
For example, doing `near` with no arguments will prompt up the above messages:
```
To bar or not to?: true
To qux or not to?: true
```
The fully written out equivalent would be: `near true qux true`

If we want to pass down dependencies from one struct to the next, we can impl the `BuilderFrom` trait.
```rust
impl BuilderFrom<TopLevel> for MySubcommand {
    fn builder_from(prev: &<TopLevel as Scoped>::Scope) -> Self::Builder {
        Self::Builder {
             qux: Some(!prev.bar),
        } 
    }
}
```
and also enable the option within the struct itself
> #[derive(Debug, Eclap)]
> #[eclap(enable(builder_from))]
> struct MySubcommand {


ChaoticTempest:
expanded code for `cli-transfer/src/main.rs`: https://gist.github.com/ChaoticTempest/8ba1d895271320ac3e1e86b656da25c9

frol:
> If we want to pass down dependencies from one struct to the next, we can impl the BuilderFrom trait.

The reason I had context in my design is to enable passing dependencies down the stream more than 1 level down. For example, we pass network selection (`connection_config`) from almost top level down to the very last step where we actually make the RPC call, and in between we also use it to validate user input (e.g. if account exists on the network or if the account has enough tokens to transfer)

ChaoticTempest:
@frol there's a huge challenge with doing this at multiple levels. Macros only act on the text we give it, so for instance in the following, we only have A and B available to us as info:
```rust
#[derive(Eclap)]
struct A {
   value: bool,

   #[eclap(single, subcommand)]
   b: B
}
```
So we can't pass down this info nicely. That's why we have to do this explicitly like with the following:
```rust
impl BuilderFrom<A> for B {
   fn builder_from(prev: &Scope) -> B::Builder {
      B::Builder { value: Some(prev.value) }
   } 
}

#[derive(Eclap)]
#[eclap(enable(builder_from)]
struct B {
   #[eclap(skip)]
   value: bool,

   #[eclap(single, subcommand)]
   c: C
}

impl BuilderFrom<B> for C {
   fn builder_from(prev: &Scope) -> B::Builder {
      B::Builder { value: Some(prev.value) }
   } 
}

#[derive(Eclap)]
#[eclap(enable(builder_from)]
struct C {
    value: bool,
}
```
This will pass down the value, and let us not have to worry about grandparents, which can be a huge problem later on if the structure changes

frol:
@ChaoticTempest I feel `B::Builder` will fall apart when `B` will have not only the skip/subcommand fields, but also long-options, and finally it will break when you would need to pass something over through `enum` variants. Context design passed these tests.

ChaoticTempest:
@frol hmm, I'm not sure I follow. How would it fall apart? `B::Builder` is pretty dynamic. It lets us fill its fields from multiple locations. The code also currently does pass thing dependencies through enum variants by default with its codegen, so that works already. But what does Context solve that Builder doesn't? They're pretty similar just that Builder is per struct

frol:
I might miss something, but I don't understand how to pass parent fields down to the children through enum:

```rust
...

struct Receiver {
    #[eclap(skip)]
    network: String, // "mainnet" | "testnet" | ....
    #[eclap(skip)]
    sender_account_id: AccountId,
    #[eclap(skip)]
    ...

    receiver_account_id: AccountId,
    #[eclap(subcommand)]
    sign: Sign,
}

enum Sign {
    SignWithKeychain(SignWithKeychainCommand),
    SignWithLedger(SignWithLedgerCommand),
}

struct SignWithKeychainCommand {
    // So, how do the fields pass over from the Receiver through Sign enum to this struct?
    #[eclap(skip)]
    network: String, // "mainnet" | "testnet" | ....
    #[eclap(skip)]
    sender_account_id: AccountId,
    #[eclap(skip)]
    ...

}
```

ChaoticTempest:
@frol ahh I see. It already does codegen internally for enums with single variant, but I didn't get to the point where I could add it for multiple variants. It'd end up looking like this for multiple:
```
#[derive(Eclap)]
#[eclap(enable(builder_from = A))]
enum B {
   C1(C1),
   C2(C2),
}

impl BuilderFrom<A> for C1 {
   ...
}

#[derive(Eclap)]
#[eclap(enable(builder_from))]
struct C1 {
   ...
}
```
There's another alternative where we don't need to specify the `#[eclap(enable(builder_from = A))]` in `B`: use the `Any` type for runtime dispatch

frol:
@ChaoticTempest We currently have the following:

```rust
enum SelectServer {
   Mainnet(ServerCommand),
   Testnet(ServerCommand),
   Betanet(ServerCommand),
   Custom(CustomServerCommand),
}

struct ServerCommand {
    #[(subcommand)]
    sender: SenderSubcommand,
}

struct CustomServerCommand {
    url: url::Url,
    #[(subcommand)]
    sender: SenderSubcommand,
}
```

I cannot think of how to deal with it using BuilderFrom

ChaoticTempest:
@frol I still need to update this branch w/ the generated code for multiple Enum variants, but what it'd look for your example would be this:
```rust
#[derive(Eclap)]
#[eclap(enable(builder_from = ParentCmd))]
enum SelectServer {
   Mainnet(ServerCommand),
   Testnet(ServerCommand),
   Betanet(ServerCommand),
   Custom(CustomServerCommand),
}

impl BuilderFrom<ParentCmd> for ServerCommand {
     fn builder_from(prev: UpAboveCmd::Scope) -> Self::Builder { Default::default() }
}

#[derive(Eclap)]
#[eclap(enable(builder_from))]
struct ServerCommand {
    #[(subcommand)]
    sender: SenderSubcommand,
}

impl BuilderFrom<ParentCmd> for CustomServerCommand {
     fn builder_from(prev: UpAboveCmd::Scope) -> Self::Builder {
          Self::Builder {
                url: Some(Url::parse(prev.url_str))
          }
     }
}

#[derive(Eclap)]
#[eclap(enable(builder_from))]
struct CustomServerCommand {
    url: url::Url,
    #[(subcommand)]
    sender: SenderSubcommand,
}
```

The generated code for the enum variant would look like this. Note, this assumes we're going to use strum underneath:
```rust
impl EnumBuilderFrom<ParentCmd> for SelectServer  {
    fn builder_from(prev: &ParentCmd::Scope) -> Self {
        match near_cli_visual::prompt_variant::<SelectServerDiscriminants>("What is your network") {
            SelectServerDiscriminants::Mainnet => {
                let builder = <ServerCommand as near_cli_visual::types::BuilderFrom<ParentCmd>>::builder_from(&scope);
                let variant = <ServerCommand as near_cli_visual::types::Build>::build(None, builder)?;
                Self::Mainnet(variant)
            }
            ...
        }
    }
}
```
Also, let me know if you want to sync up next week about it

frol:
Yeah, let's sync up next week :+1:

frol:
Just for the reference, @FroVolod implemented [interactive-clap crate](https://github.com/FroVolod/interactive-clap), and refactored near-cli-rs using it (#80), so this PR can be closed.

PR #78, opened 2021-12-08
Ci/add windows (#1)

frol:


frol:
Ooops, I merged it into upstream instead of my own fork :facepalm: 

Well, it was mostly ready, so I won't revert it, and will potentially open another small PR to address remaining problems

PR #80, opened 2021-12-24
Interactive clap master

FroVolod:


frol (approved):
Since this is a pure refactoring which dropped almost all the boilerplate we had otherwise, I will merge it into master, and we will iterate from there.

PR #84, opened 2022-02-02
Moving to near jsonrpc client rs

### LINKED ISSUE #81: Migrate to near-jsonrpc-client-rs
https://github.com/near/near-jsonrpc-client-rs

https://crates.io/crates/near-primitives
https://crates.io/crates/near-jsonrpc-client

FroVolod:
Resolves #81


src/commands/construct_transaction_command/sign_transaction/mod.rs
```diff
@@ -156,46 +155,75 @@ impl Submit {
...
+                                        near_jsonrpc_client::errors::JsonRpcServerResponseStatusError::Unexpected{status} => {
+                                            println!("JSON RPC server responded with an unexpected status code: {}", status);
+                                            return Ok(None);
+                                        }
                                     }
                                 }
-                                _ => println!("Unexpected response: {:#?}", err),
                             }
-                            return Ok(None);
+                            actix::clock::sleep(std::time::Duration::from_millis(100)).await;
```
frol, line 226: 1. This code is duplicated in `src/commands/utils_command/send_signed_transaction/mod.rs`. We should avoid copy-pasted code. 
2. This error handling breaks the regular error handling flow and thus if something fails here near-cli-rs won't exist with non-zero exit code, and so it won't be a good citizen in scripts as it will just let the failing execution passed as "executed fine". We must return an error instead of `Ok(None)`.

frol (approved):
Even though there is still room for improvement, I am merging this PR as it already provides enough value by migrating away from nearcore git dependency to a proper standalone near-jsonrpc-client-rs crate.

PR #88, opened 2022-03-09
feat: Expand ~ in path provided in interactive mode

Buckram123:
Improving user experience on interactive mode, by expanding the path. #90

### DIFF
--- a/src/commands/add_command/contract_code/contract/mod.rs
+++ b/src/commands/add_command/contract_code/contract/mod.rs
@@ -18,7 +18,7 @@ impl ContractFile {
         let input_file_path: String = Input::new()
             .with_prompt("What is a file location of the contract?")
             .interact_text()?;
-        Ok(std::path::PathBuf::from(input_file_path).into())
+        Ok(shellexpand::full(&input_file_path)?.as_ref().parse()?)
     }
 
     pub async fn process(
--- a/src/commands/add_command/contract_code/contract/mod.rs
+++ b/src/commands/add_command/contract_code/contract/mod.rs
@@ -18,11 +18,7 @@ impl ContractFile {
         let input_file_path: String = Input::new()
             .with_prompt("What is a file location of the contract?")
             .interact_text()?;
-        let file_path = if cfg!(unix) {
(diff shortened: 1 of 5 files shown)


src/commands/add_command/contract_code/contract/mod.rs
```diff
@@ -18,7 +18,7 @@ impl ContractFile {
         let input_file_path: String = Input::new()
             .with_prompt("What is a file location of the contract?")
             .interact_text()?;
-        Ok(std::path::PathBuf::from(input_file_path).into())
+        Ok(shellexpand::full(&input_file_path)?.as_ref().parse()?)
```
frol, line 21: Using `$A` variable in interactive mode is not something I would expect to work and most certainly will cause confusion if someone has a folder with `$` in the name.

I agree to expand `~`, but expanding `$A` is a stretch.

Let's use `shellexpand::tilde` instead of `full`.

frol (changes requested):

src/commands/add_command/contract_code/contract/mod.rs
Buckram123: Ok, maybe `~/$A/status_message.wasm` was a bad example, but I think user can expect something like `$WASMPATH` to be expanded to `/path/to/some.wasm`. So maybe make this feature only for unix?

frol: I would not expect variable expansion in interactive mode since the main use-case is that user will copy-paste some valid path as is. I don't use variables in interactive mode, do you?

Buckram123: I had feeling that interactive mode should act same way as cli. But if we try to add variable expansion, then it should be on every interactive input (for example $BLOCK_ID_HEIGHT, $ACCOUNT_ID...), not only on path. So yeah I agree here.

frol: @Buckram123 I feel that `~` is fine to be expanded to a home/user folder on non-unix systems. Was there a specific reason to hide it behind a feature-flag?

frol (approved):

Issue #53, opened 2021-09-02
Avoid getting balances to suggest max values in prompt


khorolets:
I was playing around the localnet and was creating an account with initial balances. Interactive prompts were suggesting values. It is localnet, so I've accepted them.

Command looked like

```
./near-cli add sub-account network custom --url http://localhost:3030/ owner-account test.near sub-account check.test.near sub-account-full-access generate-keypair deposit '999999989.9985186679375 NEAR' sign-with-keychain send
```

The result of this was

```
Transaction sent ...
Your console command:
...
```

The transaction wasn't actually sent and no error appeared. 

Decreasing the value for `deposit` made everything work as expected. The cause of the problem that suggested deposit value was an entire balance of my account. Sending transaction with this amount of tokens is impossible because nothing is even left for covering the transaction fees.

**I suggest** we stop suggesting any values for a deposit or transfer values as it might be dangerous. Let's just throw an error about "not enough funds" or something similar and let users decide what they want to do instead.

frol:
It was resolved during the migration to interactive-clap, which is included in 0.1.15 release.

PR #91, opened 2022-03-15
add sub-account and transfer don't suggest amounts and don't allow incorrect input amount

NinoLipartiia4ire:
Fixes issue https://github.com/near/near-cli-rs/issues/53.

During transfer and creating a sub-account the amounts are not suggested. If the user tries to fransfer more than allowed, the warning is shown, and the user is asked to enter the valid amount. It repeats until the correct amount is entered.


frol:
@NinoLipartiia4ire Issue #53 explicitly asks to avoid auto-population of the amount fields, which makes sense, and your PR does not address that. In fact, issue #53 was already resolved during the previous refactoring, so I will just close it.

Thanks for your contribution and interest anyway! ❤️

Issue #70, opened 2021-10-07
Show contract method names


frol:
I have just seen an issue reported on near-cli JS ([https://github.com/near/near-cli/issues/849](https://github.com/near/near-cli/issues/849)), which I believe we should be able to implement as an extension, but let's keep a tracking record here.

frol:
Previously, I thought about having a separate CLI extension for development, but recently, I changed my mind, and I actually believe that CLI is already a devtool, so we should have the developer tools in the near.cli.rs.

That said, I believe it is worth adding the method to inspect contract interface, especially when ABI is available (https://github.com/near/near-cli-rs/pull/92#issuecomment-1287698460):

```
near contract inspect "contract_account_id" network-config mainnet now
```

The expected output should at least provide the list of available function names, but if ABI is available, it should print the functions as TypeScript and/or Rust function signatures, see `cargo near abi` for reference.

vsmysle:
Working on this issue.

PR #92, opened 2022-03-15
feat: Initial version of an extension to list contract method names

Buckram123:
Creating extension for #70


extensions/method-names/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "method-names"
```
frol, line 2: It should be `near-cli-method-names` to produce a binary that `near-cli-rs` can pick up when one calls `near-cli method-names`.

frol:
@Buckram123 

1. I recommend using `interactive-clap` to build the extensions with interactivity (see #93)
2. We should not merge extensions into this repo, instead:
  2.1. setup a separate repo
  2.2. configure CI (copy `.github/workflows`)
  2.3. add the reference and a short description to near-cli-rs README

frol (changes requested):

Buckram123:
Building it [here](https://github.com/Buckram123/near-cli-method-names), working on:

- [ ] README
- [ ] Configure CI

frol:
@Buckram123 Ouch, I somehow missed your last message. Since that time devtools team developed [NEAR ABI](https://github.com/near/abi) and [`cargo near` extension](https://github.com/near/cargo-near).

I feel `cargo near` should remain dedicated to NEAR development on Rust, so we can start thinking about `near-cli-dev` which would be a generic extension for developers and yet agnostic to the language. It will include this feature with ABI support (using it to not only list the functions but also their signatures), and [more](https://github.com/near/near-cli-rs/blob/master/docs/NEW_NEAR_CLI_INTERFACE.md#developer-extension) (the commands layout still needs to be properly designed).

frol:
This is now implemented as `near contract inspect` command in #293

PR #93, opened 2022-03-15
feat: Initial version of validator extension (uses interactive clap)

frol:
* [x] use [near-jsonrpc-client-rs](https://github.com/near/near-jsonrpc-client-rs) instead of nearcore
* [x] use [interactive-clap](https://github.com/FroVolod/interactive-clap) to reduce the boilerplate
* [ ] extract it into independent repo with its own CI

Future work (in separate PRs):
* extract reusable utils out to a common crate, use it in the core project and validators extension (now validators extension is a separate project)
* simplify extensions installation (e.g. via additional commands in the core near-cli implementation)


frol:
This is now implemented in a separate repo: https://github.com/near-cli-rs/near-validator-cli-rs

Issue #42, opened 2021-08-18
Translate all in-code comments to English


volovyk-s:
This project will be developed mostly by English-speaking devs and the community. We need to translate all the comments into English.
Cyrillic characters can be found with `ag [\u0401\u0451\u0410-\u044f]`.

PR #94, opened 2022-03-15
Translated all in-code comments into English

NinoLipartiia4ire:
Fixes https://github.com/near/near-cli-rs/issues/42


frol (approved):
The comments are far from being ideal, but this PR is already an improvement since at least we converted those into English ❤️

Issue #57, opened 2021-09-02
Environment variables support


khorolets:
I think users can benefit from env support. I propose we collect a list of useful parameters that can benefit from envs.

*   `network`
     I'm testing stuff on localnet and it can be annoying to select "Custom" network all the time and provide RPC endpoint URL `[http://localhost:3030](http://localhost:3030)` every time. We could solve this by editing predefined LocalNet to the list, but assuming users might want to use their own URL it's not a great solution.

Please feel free to add other parameters you may find useful in terms of envs support

frol:
*This comment is motivated by #101*

Someone should invest their time into designing a solution to set default presets through a config file. E.g. extending a list of predefined networks with a way to configure access keys for RPC-as-a-Service (DevConsole) RPC servers, etc.

frol:
Configuration file was introduced in 0.2 version of CLI, so now you can add arbitrary number of network configurations.

@frovolod please, add "configuration" section to the README files to highlight this feature

frol:
Closing in favor of #123

PR #101, opened 2022-03-29
Environment variables support

NinoLipartiia4ire:
I suggest these changes regarding #57.

It allows creating the environment variable, which stores the URL of the custom network, for example
```
export CUSTOM_NETWORK=http://localhost:3030
```
After that the app doesn't ask to enter the URL each time the user chooses to use custom network. The value of `CUSTOM_NETWORK` is set as a default network URL if it is possible. Otherwise (if connection fails), the user is asked to enter another URL network.


frol:
Adding a single environment variable will only remedy the narrow use-case presented in #57 description, but a more general configurability is necessary.

Issue #98, opened 2022-03-21
Add keys to keychain access for OS X users


mikedotexe:
Instead of storing the private keys in unencrypted files, we can consider using something like Keychain Access in OS X. (There are other options with Linux and Windows, but this ticket is just for OS X)

It looks like there's a crate that helps connect to Apple's "Core Foundation" which allows various integrations.

Here's some general docs from Apple about this:
 [https://developer.apple.com/documentation/security/keychain_services/keychain_items/adding_a_password_to_the_keychain](https://developer.apple.com/documentation/security/keychain_services/keychain_items/adding_a_password_to_the_keychain)

Here's a crate to tap into Core Foundation:
 [https://lib.rs/crates/core-foundation](https://lib.rs/crates/core-foundation)

Here's a Golang implementation for storing keys that could be helpful:
 [https://github.com/99designs/go-keychain/blob/master/keychain.go](https://github.com/99designs/go-keychain/blob/master/keychain.go)

frol:
MacOS Keychain support is now implemented! See #122

PR #102, opened 2022-04-01
feat: Keychain access

Buckram123:
Keychain access for OS X users #98, If I did understand it correctly


frol:
@Buckram123 The issue implied that the solution should also solve the usage part of the problem (thus, not only get the keys stored, but also let those be used whenever it is applicable - signing transactions).

Also, it should be conditionally compiled for MacOS as currently this PR broke compilation on all the rest platforms (see CI)

frol:
Keep in mind that a single account can have multiple keys attached to it, and also a single near-cli-rs user might want to orchestrate multiple accounts.

frol:
FYI, I am yet to see how the created key from the keychain can be used.

Buckram123:
>  but also let those be used whenever it is applicable - signing transactions

Now signing transaction with OSX keychain also implemented, thanks.

frol:
I got really underwater with other projects. I will review this PR today

frol:
It is a great starting point, and it is good to know that we can just store the whole JSON with the secret & public keys as well as the seed phrase, etc. Yet, it is not usable, as it requires the user to specify their public key which makes it hard to use. What I suggested before is to use `~/.near-credentials` always, but store only account id, public key, and the secure storage kind or reference, so with the keychain, it could look like:

```json
{
  "account_id": "frol.near",
  "public_key": "ed25519:...",
  "storage_kind": "macos_keychain"
}
```

Then you won't need to bother users asking their public key

frol:
This PR was superseeded by #122 and MacOS keychain support is already merged into master

@Buckram123 Thank you for paving the road for this feature!

PR #110, opened 2022-05-30
chore: update near dependencies

miraclx:
Updating `nearcore` dependencies to `0.14.0` restores compatibility with parsing server responses after a breaking change was introduced.

https://github.com/near/nearcore/pull/6918
https://github.com/near/near-jsonrpc-client-rs/pull/101


miraclx:
CI tests shouldn't pass now, at least not until https://github.com/near/nearcore/pull/6918 and https://github.com/near/near-jsonrpc-client-rs/pull/100 are merged.

frol (approved):

Issue #114, opened 2022-09-28
Allow passing binary and lengthy data as function call arguments


frol:
Currently, you can only pass Unicode text as function call arguments (we use `String`), but sometimes there is a need to pass a lengthy JSON or even WASM file (e.g. upgrading the code in factory contracts). I suggest we enable these cases by asking people to choose between:

```
> How do you want to pass the function call arguments?
* json-args    - valid JSON arguments (e.g. {"token_id": "42"})
* text-args    - arbitrary text arguments
* base64-args  - Base64-encoded string (e.g. e30=)
* file-args    - Read from file (e.g. reusable JSON or binary data)
```

The final command could look like this:

```
./near-cli contract \
    call-function \
    as-transaction \
    turbo.volodymyr.testnet \
    'rate'
    json-args '{"other_user":"volodymyr.testnet", "vote":5}' \
        --prepaid-gas '3 Tgas' \
        --attached-deposit '1 NEAR' \
    fro_volod.testnet \
    network testnet \
    sign-with-keychain \
    send
```

Yet, `--prepaid-gas` and `--attached-deposit` look odd this way, so I am considering turning them into subcommand as well, and add `sign-as` named argument for the signer id, so it is not unnamed there, and rename `network` to `network-config` as there could be more than one network config using different RPCs to a single NEAR network:

```
./near-cli contract \
    call-function \
    as-transaction 'turbo.volodymyr.testnet' 'rate'
    json-args '{"other_user":"volodymyr.testnet", "vote":5}' \
    prepaid-gas '3 Tgas' \
    attached-deposit '1 NEAR' \
    sign-as fro_volod.testnet \
    network-config pagoda-testnet \
    sign-with-keychain \
    send
```

agileurbanite:
👍 shouldn't this be supported in the node version as well? https://github.com/near/near-cli

frol:
Feel free to submit a feature request there. I'm not sure about their path forward

frol:
Resolved in #117

frol:
@agileurbanite [0.2.0-pre.2 release](https://github.com/near/near-cli-rs/releases/tag/0.2.0-pre.2) now allows users to select how to pass function arguments (plaintext JSON, base64-encoded string, or read from file)

PR #117, opened 2022-10-06
Allow passing binary and lengthy data as function call arguments

FroVolod:
This is based on the proposal #114


src/commands/contract/call_function/as_transaction/mod.rs
```diff
@@ -1,30 +1,71 @@
...
+
+    pub async fn process(&self, config: crate::config::Config) -> crate::CliResult {
+        let function_args: Vec<u8> = match self.function_args_type {
+            super::call_function_args_type::FunctionArgsType::FileArgs => {
+                let data_path = std::path::PathBuf::from(self.function_args.clone());
+                std::fs::read(data_path).map_err(|err| {
+                    color_eyre::Report::msg(format!("Data file access not found! Error: {}", err))
+                })?
+            }
+            _ => self.function_args.clone().into_bytes(),
```
frol, line 45: * [ ] `JsonArgs` should be ideally validated to be valid JSON input (the code should attempt to covert `serde_json::Value::from_str()` and check the `Result`)
* [ ] `Base64Args` should be decoded (`base64::decode(self.function_args)`)

frol (approved):

PR #146, opened 2022-12-15
Added ability to import account from seed phrase or private key

FroVolod:
Implements proposal #138


src/commands/account/mod.rs
```diff
@@ -33,11 +33,9 @@ pub enum AccountActions {
     ))]
     /// View properties for an account
     ViewAccountSummary(self::view_account_summary::ViewAccountSummary),
-    #[strum_discriminants(strum(
-        message = "import-account          - Import existing account (a.k.a. \"sign in\")"
-    ))]
-    /// Import existing account (a.k.a. "sign in")
-    ImportAccount(self::import_account::Login),
+    #[strum_discriminants(strum(message = "import-account          - Import existing account"))]
+    /// Import existing account
```
frol, line 37: Let's keep the original message that has "sign in" as "import account" is not immediately obvious to the users.

src/commands/account/import_account/mod.rs
```diff
@@ -1,42 +1,64 @@
...
+    #[strum_discriminants(strum(
+        message = "import-account-from-seed-phrase         - Import existing account from seed phrase"
+    ))]
+    /// Import existing account from seed phrase
+    ImportAccountFromSeedPhrase(self::from_seed_phrase::LoginFromSeedPhrase),
+    #[strum_discriminants(strum(
+        message = "import-account-from-private-key         - Import existing account from private key"
+    ))]
+    /// Import existing account from private key
+    ImportAccountFromPrivateKey(self::from_private_key::LoginFromPrivateKey),
```
frol, line 41: Let's shorten the commands and also use "using" instead of "from" as it reads more naturally this way:

```suggestion
        message = "using-web-wallet          - Import existing account using NEAR Wallet (a.k.a. \"sign in\")"
    ))]
    /// Import existing account using NEAR Wallet (a.k.a. "sign in")
    UsingWebWallet(self::using_web_wallet::LoginFromWebWallet),
    #[strum_discriminants(strum(
        message = "using-seed-phrase         - Import existing account using a seed phrase"
    ))]
    /// Import existing account using a seed phrase
    UsingSeedPhrase(self::using_seed_phrase::LoginFromSeedPhrase),
    #[strum_discriminants(strum(
        message = "using-private-key         - Import existing account using a private key"
    ))]
    /// Import existing account using a private key
    UsingPrivateKey(self::using_private_key::LoginFromPrivateKey),
```

src/commands/account/mod.rs
FroVolod: done

frol (approved):

PR #162, opened 2023-01-21
feat: improved account view-account-summary method

frolvanya:
Changed output of view-account-summary method using prettytable-rs

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,8 +47,8 @@ cargo-util = "0.1.1"
 
 color-eyre = "0.6"
 
+bytesize = "1.1.0"
 prettytable = "0.10.0"
-byte-unit = "4.0.18"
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -47,6 +47,9 @@ cargo-util = "0.1.1"
 
 color-eyre = "0.6"
 
+prettytable = "0.10.0"
+byte-unit = "4.0.18"
```
frol, line 51: Let's use `bytesize` crate as it seems to be much slimmer and it is the crate that is used in `cargo`.

frol (approved):

frol:
@frolvanya Thanks for your contribution!

Just for reference, here is how the view-account-summary looked before:

image

After:

image

PR #165, opened 2023-03-10
feat: added self update for near-cli-rs

frolvanya:
Implemented self-update feature using `self_update` crate. Works only if `self-update` feature flag is enabled


frol (approved):
@frolvanya Great job! I tested it on MacOS (arm64 / M1-chip) and it works fine. Can you confirm whether it works on Linux and Windows?

frolvanya:
@frol Yep, it works on Windows (x86_64-pc-windows-msvc) and macOS (aarch64-apple-darwin). However, I wasn't able to cross-compile the Linux version. Let's assume that it works on Linux as well

image

image

PR #167, opened 2023-03-29
Implementation of context data processing

FroVolod:


src/main.rs
```diff
@@ -85,115 +78,32 @@ fn main() -> CliResult {
...
-            cmd,
-            Cmd {
-                top_level: crate::commands::TopLevelCommand::Extensions(
-                    crate::commands::extensions::ExtensionsCommands {
-                        extensions_actions:
-                            crate::commands::extensions::ExtensionsActions::SelfUpdate(
-                                crate::commands::extensions::self_update::SelfUpdateCommand,
-                            ),
-                    },
-                ),
```
frol, line 138: Please, don't remove `self-update` functionality that was recently added in #165.

FroVolod: done

frol (approved):
This is a massive refactoring that unblocks us on providing a better interactive experience to the CLI users!

There is still room for improvement in terms of the amount of boilerplate that we can hide in interactive-clap (e.g. strum messages), and I am still looking at how to make the whole CLI codebase leaner, but it is not blocking us from merging this PR

PR #180, opened 2023-04-16
chore: update release-plz-action to v0.5

MarcoIeni:
The latest release of release-plz removes the `git_release` github action argument.
I disabled github releases with the config file 👍

### DIFF
--- a/.github/workflows/release-plz.yml
+++ b/.github/workflows/release-plz.yml
@@ -24,9 +24,7 @@ jobs:
         if: runner.os == 'Linux'
         run: sudo apt-get update && sudo apt-get install --assume-yes libudev-dev
       - name: Run release-plz
-        uses: MarcoIeni/release-plz-action@v0.4
-        with:
-          git_release: false
+        uses: MarcoIeni/release-plz-action@v0.5
         env:
           # https://marcoieni.github.io/release-plz/github-action.html#triggering-further-workflow-runs
           GITHUB_TOKEN: ${{ secrets.CUSTOM_GITHUB_TOKEN }}
--- a/release-plz.toml
+++ b/release-plz.toml
@@ -1,0 +1,2 @@
+[workspace]
+git_release_enable = false


frol (approved):
@MarcoIeni Thanks!

Quick question regarding ["Triggering further workflow runs"](https://marcoieni.github.io/release-plz/github/trigger.html): I use a custom GITHUB_TOKEN, and this solved the issue of triggering CI jobs for the PRs created by Release-plz, but it seems that it does not solve the issue for the created tags (there is a `release.yml` pipeline, which should be triggered when a new tag appears, and the last time I cut the release, I had to remove and manually push the tag in order to trigger the job) - does 0.5 version solve it? Can Release-plz push the tag using GITHUB_TOKEN?

MarcoIeni:
Hi Vlad 👋 

> does 0.5 version solve it?

no, I haven't changed anything related to that

> I use a custom GITHUB_TOKEN

Which type is it? Classic or Fine grained?

image

I only tested the classic one so far. See https://github.com/MarcoIeni/release-plz/issues/586 which is still opened. I was planning to address this soon and test release-plz with the various types of tokens.

> and the last time I cut the release, I had to remove and manually push the tag in order to trigger the job

This sucks. And I want to clarify and solve the issue you have so that we can make it clear in the docs.
I use a GitHub app for release-plz and I'm able to trigger other workflows (I build the binaries in CI).
As a temporary workaround, you can run only the release-pr command instead of both release and release-pr in CI (docs [here](https://marcoieni.github.io/release-plz/github/index.html#example-release-pr)), and run `release-plz release --git-token ${GITHUB_TOKEN}` from the command line locally, from the `main` branch, when you want to release a new version.

frol:
@MarcoIeni it is Classic

MarcoIeni:
Can you create an [issue](https://github.com/MarcoIeni/release-plz/issues) documenting which permissions the classic token has?
I'll try to test this in another repository and reproduce the issue.
Then I'll document my learnings. 👍

Issue #186, opened 2023-04-22
call function with file args shows wrong log


hanakannzashi:
When call a function with `file-args` such as wasm code, it will show that args is `null` in console. (only log error, function call itself is fine)

<img width="419" alt="截屏2023-04-22 13 12 23" src="https://user-images.githubusercontent.com/95295094/233763983-07867551-8781-4e58-8d13-9df6d802b7b2.png">

frol:
Oh, I guess it is a display bug. I would display “<non-JSON data (X kb)>”.

@hanakannzashi Would you be open to submit a PR with the fix? I believe we need to refactor this line: https://github.com/near/near-cli-rs/blob/b34e0ce744fc43932ef3bf359c4997f2da92add9/src/common.rs#L743, currently .unwrap_or_default() leads to `null`

Chubik:
I did https://github.com/near/near-cli-rs/pull/187

PR #187, opened 2023-04-24
fixed issue call function with file args shows wrong log

Chubik:
This PR resolves https://github.com/near/near-cli-rs/issues/186 by replacing the default unwrap function with a logical alternative and adding a formatted error message for better clarity.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -738,17 +738,19 @@ pub fn print_unsigned_transaction(transaction: &crate::commands::PrepopulatedTra
                     "{:>18} {:<13} {}",
                     "",
                     "args:",
-                    serde_json::to_string_pretty(
-                        &serde_json::from_slice::<serde_json::Value>(&function_call_action.args)
-                            .unwrap_or_else(|_| {
-                                serde_json::Value::String(format!(
-                                    "<non-printable data ({} bytes)>",
-                                    function_call_action.args.len()
-                                ))
-                            }),
-                    )
-                    .unwrap_or_else(|_| "".to_string())
-                    .replace('\n', "\n                                 ")
+                    match serde_json::from_slice::<serde_json::Value>(&function_call_action.args) {
+                        Ok(parsed_args) => {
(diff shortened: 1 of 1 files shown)


frol:
@Chubik Thanks for the contribution! I feel it might be confusing to the users to see the quoted string, wouldn't it?

Check the output with the following command:

```
cargo run -- contract call-function as-transaction mycontract myfunction text-args text prepaid-gas '100.000 TeraGas' attached-deposit '0 NEAR' sign-as signer network-config testnet
```

Also, please, use `bytesize` crate (it is already used in this CLI) to display the human-readable sizes (mb, kb, b) 🙏

Chubik:
> @Chubik Thanks for the contribution! I feel it might be confusing to the users to see the quoted string, wouldn't it?
> 
> Check the output with the following command:
> 
> ```
> cargo run -- contract call-function as-transaction mycontract myfunction text-args text prepaid-gas '100.000 TeraGas' attached-deposit '0 NEAR' sign-as signer network-config testnet
> ```
> 
> Also, please, use `bytesize` crate (it is already used in this CLI) to display the human-readable sizes (mb, kb, b) 🙏

Thank you for your feedback! I understand your concern about the potential confusion regarding the quoted string :)
I have tested the output using the command provided:

`cargo run -- contract call-function as-transaction mycontract myfunction text-args text prepaid-gas '100.000 TeraGas' attached-deposit '0 NEAR' sign-as signer network-config testnet`

output:

`
actions:
   -- function call:
                   method name:  myfunction
                   args:         <non-printable data (4 B)>
                   gas:          100.000 TeraGas
                   deposit:      0 NEAR
`

I will also make sure to use the bytesize crate, which is already utilized in this CLI, to display the human-readable sizes (MB, KB, B) for better clarity. I did the necessary adjustments and updated the PR accordingly.

frol:
@Chubik Thanks again for the quick fix! I merged it and will release it later this week.

Issue #205, opened 2023-05-29
Delete several keys at once


frol:
Currently, `near account delete-key` only allows the deletion of one key at a time, but when someone wants to clean up their account from keys, they usually want to delete many at once (if not all).

It would be great to allow a comma-separated list of public keys in CLI args and a checkbox list of access keys for the given account in interactive mode.

Given that when we ask for the access key, the user has not yet specified the network, let's query all the networks for the access keys for the given account id.

PR #207, opened 2023-05-30
feat: multi public-key for account delete-key command

Mehrbod2002:
Fixes : #205 
multi public-key for account delete-key command separated  by `,` .

### DIFF
--- a/src/commands/account/delete_key/mod.rs
+++ b/src/commands/account/delete_key/mod.rs
@@ -26,20 +26,20 @@ impl DeleteKeyCommandContext {
         Ok(Self {
             config: previous_context.0,
             owner_account_id: scope.owner_account_id.clone().into(),
-            public_keys: scope.public_keys.clone().0,
+            public_keys: scope.public_keys.clone().into(),
         })
     }
 }
 
 impl From<DeleteKeyCommandContext> for crate::commands::ActionContext {
     fn from(item: DeleteKeyCommandContext) -> Self {
-        let public_keys = item.public_keys.clone();
         let on_after_getting_network_callback: crate::commands::OnAfterGettingNetworkCallback =
             std::sync::Arc::new(move |_network_config| {
                 Ok(crate::commands::PrepopulatedTransaction {
                     signer_id: item.owner_account_id.clone(),
                     receiver_id: item.owner_account_id.clone(),
-                    actions: public_keys
+                    actions: item
+                        .public_keys
                         .clone()
                         .into_iter()
                         .map(|public_key| {
--- a/src/commands/account/delete_key/mod.rs
+++ b/src/commands/account/delete_key/mod.rs
@@ -6,8 +6,6 @@ mod use_publickeylist_type;
 pub struct DeleteKeyCommand {
     /// Which account should you delete the access key for?
     owner_account_id: crate::types::account_id::AccountId,
-    #[interactive_clap(long)]
-    #[interactive_clap(skip_default_from_cli_arg)]
     /// Enter the public keys you wish to delete (separated by comma):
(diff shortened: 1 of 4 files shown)


frol:
@Mehrbod2002 Thanks for the contribution!

@FroVolod Please, review this PR. It seems that it does not have checkbox-list in the interactive mode and instead asks a user to specify public keys as a comma-separated list, but it might be already good enough for the time being while checkbox-list can be implemented later.

src/commands/account/delete_key/use_publickeylist_type/mod.rs
```diff
@@ -0,0 +1,32 @@
+use interactive_clap::ToCli;
+
+#[derive(Debug, Clone)]
+pub struct PublicKeyList(pub Vec<crate::types::public_key::PublicKey>);
```
frol, line 4: I think you can use `near_crypto::PublicKey` here directly. There is no point in using the wrapper type. And then you don't need to repackage the items at all: https://github.com/near/near-cli-rs/pull/207/files#diff-a637fdd677d48fb8d9a4460b98e3ed6a82e647bebbf16435e42e1e667f829ef0R35-R37

src/commands/account/delete_key/mod.rs
```diff
@@ -26,23 +28,36 @@ impl DeleteKeyCommandContext {
...
             config: previous_context.0,
             owner_account_id: scope.owner_account_id.clone().into(),
-            public_key: scope.public_key.clone().into(),
+            public_keys: scope
+                .public_keys
+                .clone()
+                .0
+                .into_iter()
+                .map(Into::into)
+                .collect(),
```
frol, line 37: See another comment.

```suggestion
```

src/commands/account/delete_key/mod.rs
```diff
@@ -26,23 +28,36 @@ impl DeleteKeyCommandContext {
...
         })
     }
 }
 
 impl From<DeleteKeyCommandContext> for crate::commands::ActionContext {
     fn from(item: DeleteKeyCommandContext) -> Self {
+        let public_keys_clone = std::sync::Arc::new(item.public_keys.clone());
```
frol, line 44: Drop the Arc, cloning is enough.

```suggestion
        let public_keys = item.public_keys.clone();
```

frol (changes requested):

src/commands/account/delete_key/use_publickeylist_type/mod.rs
```diff
@@ -0,0 +1,32 @@
+use interactive_clap::ToCli;
```
frol, line 1: Please, move this file to `src/types/public_key_list.rs`

src/commands/account/delete_key/mod.rs
```diff
@@ -26,23 +28,36 @@ impl DeleteKeyCommandContext {
...
-                            public_key: item.public_key.clone(),
-                        },
-                    )],
+                    actions: public_keys_clone
+                        .clone()
+                        .iter()
+                        .map(|public_key| {
+                            near_primitives::transaction::Action::DeleteKey(
+                                near_primitives::transaction::DeleteKeyAction {
+                                    public_key: public_key.clone(),
```
frol, line 56: I believe this should work and void unnecessary cloning:

```suggestion
                        .into_iter()
                        .map(|public_key| {
                            near_primitives::transaction::Action::DeleteKey(
                                near_primitives::transaction::DeleteKeyAction {
                                    public_key,
```

Mehrbod2002:
@frol Now you can review it .

frol (approved):
Looks good to me now. Waiting for CI and ready to merge!

PR #209, opened 2023-05-31
Sign offline

### LINKED ISSUE #201: Extend Transaction Signing options with `sign-offline`
Currently, all the [available singing options](https://github.com/near/near-cli-rs/blob/48cb8fcabfb277012a8349d08c649037cee39e5e/src/transaction_signature_options/mod.rs#L21-L53) require a user to have their access key on the same machine as they prepare the transaction. I want to be able to use an air-gapped machine to sign my transactions.

This is the flow:

1. [online or offline] Prepare a transaction - select `sign-offline` to get an unsigned transaction [this issue] with empty-values instead of `public_key`, `nonce`, and `block_hash`
2. [offline] Sign the prepared transaction by combining together:
  a. [offline] The prepared unsigned transaction (we can use base64 encoding to simplify copy-pasting)
  b. [offline] The `public_key` of the access key that corresponds to the `signer_id` of the transaction
  c. [online] The `nonce` for the access key and a recent `block_hash` should be obtained from a computer that has access to NEAR network (usually connected to the Internet)
3. [online] Submit the signed transaction from a computer that has access to NEAR network (see #200)

As of today, near-cli-rs does this all in one go, and only allows users to opt-out of sending a signed transaction immediately (effectively, you can only do (1) and (2) together), and while near-cli-rs can be used in air-gapped environment to prepare **and** sign a transaction, it would be great to have more granular control.

### LINKED ISSUE #202: Respect `--nonce` and `--block-hash` arguments to allow signing transactions in offline
Currently, `--nonce` and `--block-hash` CLI arguments seem to be ignored in all the signing options.

https://github.com/near/near-cli-rs/blob/48cb8fcabfb277012a8349d08c649037cee39e5e/src/transaction_signature_options/sign_with_access_key_file/mod.rs#L15-L18

https://github.com/near/near-cli-rs/blob/48cb8fcabfb277012a8349d08c649037cee39e5e/src/transaction_signature_options/sign_with_access_key_file/mod.rs#L59-L62

near-cli-rs should not make any calls to RPC when `nonce` and `block-hash` is specified, so it will work on computers that are not connected to the Internet.

FroVolod:
Resolves #201 
Resolves #202


src/transaction_signature_options/sign_with_private_key/mod.rs
```diff
@@ -19,7 +20,10 @@ pub struct SignPrivateKey {
     pub nonce: Option<u64>,
     #[interactive_clap(long)]
     #[interactive_clap(skip_default_input_arg)]
-    pub block_hash: Option<String>,
+    pub block_hash: Option<crate::types::crypto_hash::CryptoHash>,
+    #[interactive_clap(long)]
+    #[interactive_clap(skip_default_input_arg)]
+    pub block_height: Option<u64>,
```
frol, line 26: Use `near_primitives::types::BlockHeight` instead of primitive `u64` type for `block_height`.

Also, update it below in `input_block_height`

src/transaction_signature_options/sign_with_private_key/mod.rs
```diff
@@ -46,25 +51,38 @@ impl SignPrivateKeyContext {
...
-            .json_rpc_client()
-            .blocking_call_view_access_key(
-                &previous_context.prepopulated_transaction.signer_id,
-                &public_key,
-                near_primitives::types::BlockReference::latest(),
+        let (nonce, block_hash, block_height) = if previous_context.offline {
+            (
+                scope.nonce.unwrap(),
+                scope.block_hash.unwrap().0,
+                scope.block_height.unwrap(),
```
frol, line 58: Return `Err` instead of crashing the application.

```rust
scope.nonce.wrap_err("Nonce is required to sign a transaction in offline mode")?,
```

src/types/mod.rs
```diff
@@ -1,5 +1,6 @@
 pub mod account_id;
 pub mod api_key;
+pub mod bool;
```
frol, line 3: update `interactive-clap` to 0.2.4 and remove this type.

src/transaction_signature_options/sign_with_keychain/mod.rs
```diff
@@ -218,6 +258,47 @@ impl interactive_clap::FromCli for SignKeychain {
     {
         let mut clap_variant = optional_clap_variant.unwrap_or_default();
 
+        let signer_public_key: crate::types::public_key::PublicKey = if context.offline {
```
frol, line 261: Make `signer_public_key` as an optional parameter like nonce/block_hash/block_height, and move this code to the `input_signer_public_key` function.

frol (changes requested):

src/types/mod.rs
FroVolod: Done

src/transaction_signature_options/sign_with_private_key/mod.rs
FroVolod: Done

src/transaction_signature_options/sign_with_private_key/mod.rs
FroVolod: Done

src/transaction_signature_options/sign_with_keychain/mod.rs
FroVolod: Done

src/commands/transaction/construct_transaction/add_action_1/add_action/delete_account/mod.rs
```diff
@@ -37,3 +40,53 @@ impl From<DeleteAccountActionContext> for super::super::super::ConstructTransact
...
+                account_id
+            )]
+            Yes {
+                account_id: crate::types::account_id::AccountId,
+            },
+            #[display(fmt = "No, I know this account exists and want to continue.")]
+            No,
+        }
+        let select_choose_input =
+            Select::new("\nDo you want to check the existence of the specified account so that you don’t waste tokens with sending a transaction that won't succeed?",
```
frol, line 68: Apply the same suggestions for delete-account command that I requested above to all the add_action_* modules.

src/commands/transaction/send_meta_transaction/sign_as/mod.rs
```diff
@@ -73,11 +73,16 @@ impl RelayerAccountId {
     fn input_relayer_account_id(
         context: &super::SendMetaTransactionContext,
     ) -> color_eyre::eyre::Result<Option<crate::types::account_id::AccountId>> {
+        let mut relayer_account_id: crate::types::account_id::AccountId =
+            CustomType::new("What is the relayer account ID?").prompt()?;
+
+        if context.global_context.offline {
+            return Ok(Some(relayer_account_id));
+        }
```
frol, line 81: Let's move the prompt back into the loop. It is not a problem that `offline` would be checked in every loop instance, but copy-pasted message asking relayer account id is easy to miss and have two different messages in the future.

src/main.rs
```diff
@@ -118,6 +147,7 @@ fn main() -> crate::common::CliResult {
     if !matches!(
         cli_cmd,
         Ok(Some(CliCmd {
+            offline: _,
```
frol, line 150: We don't need to match all the specific fields, so we can just use `..` here (probably move after `top_level` field):

```suggestion
            ..
```

src/main.rs
```diff
@@ -4,27 +4,52 @@ use clap::Parser;
...
 
 pub use common::CliResult;
-pub type GlobalContext = (crate::config::Config,);
+pub use near_cli_rs::GlobalContext;
 
 #[derive(Debug, Clone, interactive_clap::InteractiveClap)]
-#[interactive_clap(context = crate::GlobalContext)]
-struct Cmd {
+#[interactive_clap(input_context = crate::GlobalContext)]
+#[interactive_clap(output_context = CmdContext)]
```
frol, line 23: Instead of `GlobalContext` -> `CmdContext` and then later back to `GlobalContext` (through .into() implementation), make it the other way around: `input_context = ConfigContext, output_context = GlobalContext`, and `ConfigContext` would only hold `config` while `GlobalContext` will add `offline` to it.

src/transaction_signature_options/mod.rs
```diff
@@ -21,7 +22,7 @@ pub const META_TRANSACTION_VALID_FOR_DEFAULT: u64 = 1000;
 pub enum SignWith {
     #[cfg(target_os = "macos")]
     #[strum_discriminants(strum(
-        message = "sign-with-macos-keychain         - Sign the transaction with a key saved in macOS keychain"
+        message = "sign-with-macos-keychain         - Sign the transaction with a key saved in macOS keychain (Online only!)"
```
frol, line 25: We agreed that we don't need to confuse users with "online mode" as it is rare case when someone would want to use near-cli-rs in offline mode, so revert this change (remove the "(Online only!)" note)

frol (changes requested):
Also, address the clippy warning in [`src/network/mod.rs`](https://github.com/near/near-cli-rs/pull/209/files#diff-bc565c9df4b377f07d29b6132b181f1a239dfd8acab3700841961ed4550f8975)

src/transaction_signature_options/mod.rs
FroVolod: Done

src/commands/transaction/construct_transaction/add_action_1/add_action/delete_account/mod.rs
FroVolod: Done

src/commands/transaction/send_meta_transaction/sign_as/mod.rs
FroVolod: Done

frol (approved):

PR #215, opened 2023-06-05
Sign base64 encoding transaction

FroVolod:


frol (approved):
This new command looks great 👍 

* [x] It would be great to mention this command in `sign-later` documentation and/or in the result output
* [x] It would be great to mention `send-signed-transaction` command in the `display` command when the transaction is not sent immediately
* [x] Add this new command to the GUIDE

PR #218, opened 2023-06-08
feat: Allow to specify a custom Web Wallet URL when importing account (default to MyNearWallet)

FroVolod:


docs/GUIDE.en.md
```diff
@@ -243,7 +243,17 @@ near account \
...
 
-You will be redirected to the browser for authorization. After successful authorization in _[NEAR Wallet](https://wallet.near.org/)_, you need to return to the terminal and enter your login.
+You will be redirected to the browser for authorization.  
+Default wallet url is https://app.mynearwallet.com/ (for testnet - https://testnet.mynearwallet.com/). But if you want to change to a different wallet url, you can use `--walletUrl` option:
+```txt
+near account \
+    import-account \
+    using-web-wallet \
+    --wallet-url 'https://wallet.testnet.near.org/'\
+    network-config testnet
```
frol, line 253: Looking at this command I realized that it would be great to have ad-hoc ways to alter `network-config`, so adding optional parameters to `NetworkConfig` makes total sense, and then the command would be:

```txt
near account \
    import-account \
    using-web-wallet \
    network-config testnet \
    --wallet-url 'https://wallet.testnet.near.org/'
```

Alternatively, we may consider implementing "Wallet Selector" when instead of direct URL it will ask if one wants to use "NEAR Wallet" or "MyNearWalllet" or "Meteor Wallet" or ... and then the command would be:

```txt
near account \
    import-account \
    using-web-wallet \
    "MyNearWallet" \
    network-config testnet
```

This path with Wallet selectors is harder to implement and maintain, so I would at least start with the first option (extending `network-config` with optional parameters)

src/network/mod.rs
```diff
@@ -44,7 +57,7 @@ impl interactive_clap::FromCli for Network {
             .expect("Failed to get network config!")
             .clone();
 
-        match (context.on_after_getting_network_callback)(&network_config) {
+        match (context.on_after_getting_network_callback)(&network_config, wallet_url) {
```
frol, line 60: Instead of adding another parameter to the callback, let's just update the `network_config` when `optional_wallet_url` is Some. This way you won't need to "teach" all the handlers how to react to the `optional_wallet_url`, and they can just rely on `network_config.wallet_url`.

frol (changes requested):

src/network/mod.rs
FroVolod: Done

frol (approved):

Issue #222, opened 2023-06-23
Order networks selection based on the selected account id


frol:
Currently, when prompted to select network `near` CLI uses the order specified in the config.toml, it would be useful if we could bubble up network configs relevant to the specified account id (e.g. when `frol.testnet` is specified, we would move all the "testnet" connections up before mainnet ones).

<img width="1144" alt="image" src="https://github.com/near/near-cli-rs/assets/304265/eb690be2-0a7c-46e1-8d90-322f56e39e12">

(^ I wish that 3 last items would be at the beginning of the list)

Note that implicit account ids are not guessable, while named accounts end with `.near` suffix on mainnet, and with `.testnet` on testnet. NetworkConnection has `network_name` that can be used to identify the relevant network connection.

thaodt:
I'm working on this one.

frol:
Resolved in #225

PR #225, opened 2023-07-05
Improve commands - order networks selection based on the selected account id

thaodt:
This PR is served for #222.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -1572,38 +1572,40 @@ pub fn display_access_key_list(access_keys: &[near_primitives::views::AccessKeyI
     table.printstd();
 }
 
-/// sort network names based on account_id if provided
-/// if account_id is not provided, sort based on network name
-/// this is to make sure that the default network is always the same
-/// otherwise, sort the network names which maps to the same account_id to the top of the list
+/// Interactive prompt for network name.
+///
+/// If account_ids is provided, show the network connections that are more
+/// relevant at the top of the list.
 pub fn input_network_name(
     config: &crate::config::Config,
-    account_id: Option<&crate::types::account_id::AccountId>,
+    account_ids: &[near_primitives::types::AccountId],
 ) -> color_eyre::eyre::Result<Option<String>> {
-    let default_variants: Vec<_> = config.network_connection.keys().collect();
-
-    let variants = match account_id {
-        Some(account_id) => {
-            let account_id_str =
-                <crate::types::account_id::AccountId as AsRef<str>>::as_ref(account_id);
-            let last_prefix = account_id_str.rsplit('.').next();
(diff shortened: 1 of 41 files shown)


src/common.rs
```diff
@@ -1572,10 +1572,49 @@ pub fn display_access_key_list(access_keys: &[near_primitives::views::AccessKeyI
...
 }
 
+/// sort network names based on account_id if provided
+/// if account_id is not provided, sort based on network name
+/// this is to make sure that the default network is always the same
+/// otherwise, sort the network names which maps to the same account_id to the top of the list,
+/// and then sort the rest of the network names alphabetically
 pub fn input_network_name(
     config: &crate::config::Config,
+    account_id: Option<crate::types::account_id::AccountId>,
```
frol, line 1582: I don't see a reason to pass `account_id` by value, so I suggest passing it by reference to avoid unnecessary clones on the caller side:

```suggestion
    account_id: Option<&crate::types::account_id::AccountId>,
```

src/common.rs
```diff
@@ -1572,10 +1572,49 @@ pub fn display_access_key_list(access_keys: &[near_primitives::views::AccessKeyI
...
+                std::cmp::Ordering::Greater
+            } else {
+                a.cmp(b)
+            }
+        });
+        variants
+    } else {
+        let mut variants = config.network_connection.keys().collect::<Vec<_>>();
+        variants.sort();
+        variants
```
frol, line 1615: Preserve the configuration order, do not sort.

src/common.rs
```diff
@@ -1572,10 +1572,49 @@ pub fn display_access_key_list(access_keys: &[near_primitives::views::AccessKeyI
...
+                .clone();
+
+            if a_account_id.contains(vec_account_ids[1]) {
+                std::cmp::Ordering::Less
+            } else if b_account_id.contains(vec_account_ids[1]) {
+                std::cmp::Ordering::Greater
+            } else {
+                a.cmp(b)
+            }
+        });
```
frol, line 1610: * [ ] We don't need to sort the list, just lift up more relevant options
* [ ] Don't use `unwrap` (in this specific case it could have been avoided by using `iter()` instead of `keys()`)
* [ ] Don't use `[1]` as it may crash if account id does not have `.` (e.g. implicit account name)
* [ ] We only need the last prefix of the specified account id

frol (changes requested):
This PR does not meet the task requirements.

thaodt:
Finished per your review, please recheck.

frol (approved):
I realized that leaving the review comments would take longer than finishing it myself, so I refactored the code to always use `near_primitives::types::AccountId` instead of `crate::types::account_id::AccountId` in Context structs, and added support for multiple account ids (e.g. sender/receiver/nft_contract_account_id)

Issue #195, opened 2023-05-16
Implement system keychain support for Windows and Linux


frol:
Currently, near-cli-rs only supports macOS keychain to store access keys (it stores JSON data with private key as a password entry in the keychain and reads it in memory when user wants to sign a transaction).

There is a nice crate that has Windows, Linux, macOS, and iOS support: https://docs.rs/keyring/. I suggest we try using it and rename the “macos-keychain” commands into “keychain” and the current “keychain” into “legacy-keychain”. For backwards compatibility, the new “keychain” command should check the legacy keychain folder when signing transactions if there is no accesss key in the system keychain before returning an error.

---

_If anyone wants to work on this feature, feel free to ask any questions in the comments and I’m happy to give you pointers about where to start_

gautamprikshit1:
Can you provide with some extra pointers to get started on this issue. Thanks in advance

frol:
Here is the most relevant starting point: https://github.com/near/near-cli-rs/blob/251e45fbe6fc8ea730df4a942dc86274a27c9fe5/src/transaction_signature_options/sign_with_macos_keychain/mod.rs

`keychain` is instantiated there:

https://github.com/near/near-cli-rs/blob/251e45fbe6fc8ea730df4a942dc86274a27c9fe5/src/transaction_signature_options/sign_with_macos_keychain/mod.rs#L49C13-L49C21

and then used later.

This module is feature-gated and only available on macOS (for obvious reasons). The `keyring` crate above should be able to help us to implement similar logic, but for all the major platforms at once.

shariffdev:
@frol I put up a PR for this here https://github.com/near/near-cli-rs/pull/232

PR #226, opened 2023-07-11
feat: New command export-account

### LINKED ISSUE #220: Allow an easy way to export account
I am thinking about having `near account export-account` that would reveal the private key and seed phrase (if available) stored in the keychain.

Similarly to `near account import-account` there will be several ways to export:

* `near account export-account <account-id> using-web-wallet network-config testnet`
* `near account export-account <account-id> using-seed-phrase network-config testnet`
* `near account export-account <account-id> using-private-key network-config testnet`

For web wallet option, near CLI should [generate a URL for auto-import of access key](https://stackoverflow.com/questions/71238889/how-to-import-an-account-into-the-near-wallet-using-only-the-private-key-no-see) and open it in the browser:

```
https://WALLET_URL/auto-import-secret-key#YOUR_ACCOUNT_ID/YOUR_PRIVATE_KEY
```

For other options, just print the access key.

Given that a single account might have several access keys, let's follow the same strategy as for signing transactions (fetch the list of access keys from RPC, and export first full access key that is available locally)

FroVolod:
Resolves #220


src/commands/account/export_account/using_private_key/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+        let config = previous_context.global_context.config.clone();
+        let account_id = previous_context.account_id.clone();
+
+        let on_after_getting_network_callback: crate::network::OnAfterGettingNetworkCallback =
+            std::sync::Arc::new({
+                move |network_config| {
+                    #[cfg(target_os = "macos")]
+                    {
+                        if let Ok(account_key_pair) =
+                            super::using_web_wallet::get_account_key_pair_from_macos_keychain(
```
frol, line 27: Let's adjust this code after #195

src/commands/account/export_account/using_seed_phrase/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+        let config = previous_context.global_context.config.clone();
+        let account_id = previous_context.account_id.clone();
+
+        let on_after_getting_network_callback: crate::network::OnAfterGettingNetworkCallback =
+            std::sync::Arc::new({
+                move |network_config| {
+                    #[cfg(target_os = "macos")]
+                    {
+                        if let Ok(password_list) =
+                            super::using_web_wallet::get_password_list_from_macos_keychain(
```
frol, line 29: If you need this function in all three submodules, move it to the parent instead of keeping it in the `using_web_wallet` module.

Apply this reasoning to other similar functions.

FroVolod: Done

src/commands/account/export_account/using_private_key/mod.rs
FroVolod: Done

src/commands/account/export_account/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+pub fn get_account_key_pair_from_keychain(
+    network_config: &crate::config::NetworkConfig,
+    account_id: &near_primitives::types::AccountId,
+) -> color_eyre::eyre::Result<crate::transaction_signature_options::AccountKeyPair> {
+    let password = get_password_from_keychain(network_config, account_id)?;
+    let account_key_pair = serde_json::from_str(&password);
+    if let Ok(key_pair) = account_key_pair {
+        return Ok(key_pair);
+    }
+    Err(color_eyre::eyre::Report::msg("Error reading data"))
```
frol, line 82: I believe the following code should work:

```suggestion
    account_key_pair.wrap_err("Error reading data")
```

src/commands/account/export_account/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+                        .ok()?;
+                keyring.get_password().ok()
+            });
+
+        match res {
+            Some(password) => password,
+            None => {
+                return Err(color_eyre::Report::msg("No access keys found in keychain"));
+            }
+        }
```
frol, line 126: `wrap_err` should do the trick:

```suggestion
        access_key_list
            .keys
            .into_iter()
            .filter(|key| {
                matches!(
                    key.access_key.permission,
                    near_primitives::views::AccessKeyPermissionView::FullAccess
                )
            })
            .map(|key| key.public_key)
            .find_map(|public_key| {
                let keyring =
                    keyring::Entry::new(&service_name, &format!("{}:{}", account_id, public_key))
                        .ok()?;
                keyring.get_password().ok()
            })
            .wrap_err("No access keys found in keychain")?
```

src/commands/account/export_account/using_private_key/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+                        super::get_account_key_pair_from_keychain(network_config, &account_id)
+                    {
+                        println!(
+                            "Here is the private key for account <{}>: {}",
+                            account_id, account_key_pair.private_key,
+                        );
+                        return Ok(());
+                    }
+
+                    if let Ok(account_key_pair) = super::get_account_key_pair_from_legacy_keychain(
```
frol, line 34: `wrap_err?`

FroVolod: Done

src/commands/account/export_account/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+    for access_key in access_key_list.keys {
+        let account_public_key = access_key.public_key.to_string();
+        let is_full_access_key: bool = match &access_key.access_key.permission {
+            near_primitives::views::AccessKeyPermissionView::FullAccess => true,
+            near_primitives::views::AccessKeyPermissionView::FunctionCall {
+                allowance: _,
+                receiver_id: _,
+                method_names: _,
+            } => false,
+        };
```
frol, line 194: It seems there is no value in having `is_full_access_key` variable, as if it is `false`, we will skip any operations anyway, so we may just continue to the next iteration of the loop, and try the next access key:

```suggestion
        match &access_key.access_key.permission {
            near_primitives::views::AccessKeyPermissionView::FullAccess => {}
            near_primitives::views::AccessKeyPermissionView::FunctionCall { .. } => { continue; }
        }
```

Also, you don't need to name all the fields if you don't use them, just use `{ .. }`

src/commands/account/export_account/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+                }
+                let data =
+                    std::fs::read_to_string(&data_path).wrap_err("Access key file not found!")?;
+                if serde_json::from_str::<crate::common::KeyPairProperties>(&data).is_ok() {
+                    return Ok(data_path);
+                } else {
+                    return Err(color_eyre::eyre::Report::msg(format!(
+                        "There are no master seed phrase in keychain to export for account <{account_id}>."
+                    )));
+                }
```
frol, line 220: You can shorten it like this:

```suggestion
                serde_json::from_str::<crate::common::KeyPairProperties>(&data).wrap_err_with(|| format!(
                        "There are no master seed phrase in keychain to export for account <{account_id}>."
                    ))?;
                return Ok(data_path);
```

src/commands/account/export_account/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+            .read_dir()
+            .wrap_err("There are no access keys found in the keychain for the account.")?;
+        for entry in dir.flatten() {
+            if entry
+                .path()
+                .file_stem()
+                .unwrap()
+                .to_str()
+                .unwrap()
+                .contains(account_public_key.rsplit(':').next().unwrap())
```
frol, line 205: Instead of splitting, use replace, and do it on `account_public_key` variable initialization (above) to avoid this operation being executed for each `entry`.

frol (changes requested):

src/commands/account/export_account/mod.rs
FroVolod: Done

frol (approved):

PR #259, opened 2023-11-16
feat: ledger blind signing

### LINKED ISSUE #33: Signing a transaction with Ledger fails if the transaction is "too" big
Certain transactions like those that involve contract deployment or huge blob of arguments in a function call read to an obscure error, which was already reported here [https://github.com/near/near-cli/issues/612](https://github.com/near/near-cli/issues/612) and here https://github.com/LedgerHQ/app-near/issues/13

**There is no solution yet. Further discussion should happen in the thread mentioned above.**

dj8yfo:
resolves #33 
tmp
verify:
tmp

tx [link](https://nearblocks.io/txns/26GbdoRxMxTn4ggyStxTPHjTqciWWMka6H4TRTUrox5g) exactly from flow on screens above

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -531,7 +531,7 @@ fn compute_hash(bytes: &[u8]) -> [u8; 32] {
 }
 
 pub fn print_full_signed_transaction(transaction: near_primitives::transaction::SignedTransaction) {
-    eprintln!("{:<55} {}", "signature:", transaction.signature);
+    eprintln!("{:<25} {}\n", "signature:", transaction.signature);
     crate::common::print_full_unsigned_transaction(transaction.transaction);
 }
 
@@ -539,10 +539,17 @@ pub fn print_full_unsigned_transaction(transaction: near_primitives::transaction
     let bytes = transaction
         .try_to_vec()
         .expect("Transaction is not expected to fail on serialization");
+    eprintln!("transaction payload (without signature) hash:");
     eprintln!(
-        "{:<55} {}\n\n",
-        "transaction payload (without signature) sha256 hash:",
+        "{:<25} {}",
+        "SHA256 hash(hex):",
         hex::encode(&compute_hash(&bytes))
+    );
+
+    eprintln!(
+        "{:<25} {}\n\n",
+        "SHA256 hash(base58):",
+        CryptoHash::hash_bytes(&bytes)
     );
 
     eprintln!("{:<13} {}", "public_key:", &transaction.public_key);
--- a/src/common.rs
+++ b/src/common.rs
@@ -5,6 +5,7 @@ use std::str::FromStr;
 
 use color_eyre::eyre::WrapErr;
 use futures::{StreamExt, TryStreamExt};
+use near_primitives::borsh::BorshSerialize;
 use prettytable::Table;
 
 use near_primitives::{hash::CryptoHash, types::BlockReference, views::AccessKeyPermissionView};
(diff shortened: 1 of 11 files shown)


frol:
@dj8yfo Is there a chance to improve the error message for old Ledger App?

telegram-cloud-photo-size-2-5404517375199530626-y

src/common.rs
```diff
@@ -517,6 +519,47 @@ pub fn generate_keypair() -> color_eyre::eyre::Result<KeyPairProperties> {
...
+
+pub fn print_full_unsigned_transaction(transaction: near_primitives::transaction::Transaction) {
+    let bytes = transaction
+        .try_to_vec()
+        .expect("Transaction is not expected to fail on serialization");
+    eprintln!("transaction payload (without signature) hash:");
+    eprintln!(
+        "{:<25} {}",
+        "SHA256 hash(hex):",
+        hex::encode(&compute_hash(&bytes))
```
frol, line 546: You don't need to implement the `compute_hash` function, Transaction type has [`get_hash_and_size()` method](https://docs.rs/near-primitives/latest/near_primitives/transaction/struct.Transaction.html#method.get_hash_and_size), just use it (and remove the method and explicit `sha2` dependency.

src/common.rs
```diff
@@ -517,6 +519,47 @@ pub fn generate_keypair() -> color_eyre::eyre::Result<KeyPairProperties> {
...
+}
+
+pub fn print_full_unsigned_transaction(transaction: near_primitives::transaction::Transaction) {
+    let bytes = transaction
+        .try_to_vec()
+        .expect("Transaction is not expected to fail on serialization");
+    eprintln!("transaction payload (without signature) hash:");
+    eprintln!(
+        "{:<25} {}",
+        "SHA256 hash(hex):",
```
frol, line 545: In NEAR, transaction hash is usually represented as base58 and even though I wish we wouldn't reinvent the wheel and used hex since day one, but we did not. Unless Ledger requires and shows hex, let's use it, otherwise, I would make things simpler and only use base58 and call it "Transaction hash (Base58-encoded SHA-256 hash)".

src/transaction_signature_options/sign_with_ledger/mod.rs
```diff
@@ -44,6 +46,76 @@ pub struct SignLedgerContext {
...
+        eprintln!("{}", hex::encode(&hash.0));
+
+        eprintln!(
+            "\nUnsigned transaction (serialized as base64):\n{}\n",
+            crate::types::transaction::TransactionAsBase64::from(unsigned_transaction)
+        );
+        eprintln!("Before proceeding with blind signature,");
+        eprintln!("you have ability to verify unsigned transaction's details and exact SHA256 correspondence,");
+        eprintln!("if you have concerns with trust to current computer, where near CLI command is being run,");
+        eprintln!("on another computer(s) with the following helper command on near CLI:");
```
frol, line 75: It seems that part of the sentence is missing. Let's also print the whole sentence in a single print, so it is up to the terminal to wrap it over the lines.

frol (changes requested):

src/transaction_signature_options/sign_with_ledger/mod.rs
dj8yfo: Blind signature means that transaction is prepared by CLI, but cannot be reviewed on the Ledger device. In order to be absolutely sure that the transaction you are signing is not forged, take the constructed transaction, verify its content using NEAR CLI on another host or use any other tool capable displaying unsigned NEAR transactions, and confirm that the SHA256 hash matches the one displayed above and another identical one, that will be displayed on your Ledger device after confirming the prompt. Following helper command on near CLI can be used:

frol (approved):

PR #261, opened 2023-11-21
Show hash-to-sign when using the sign_later transaction signature option

jjambazov-bc:
In delegated signing contexts it is useful to see the SHA256 hash of the unsigned transaction when using the `sign-later` transaction signature option.


frol (approved):

frol:
@jjambazov-bc Thank you for the PR!

PR #269, opened 2023-12-02
feat: Added the ability to select access keys to remove from the list of public keys

### LINKED ISSUE #267: Enhanced key management
At the moment the function
`$ near account list-keys <account> network-config mainnet now`
displays the keys of the account in an order, presumably, from the newest to the oldest. It would be good to be able to specify listing of full access keys only or highlight somehow these keys. It is important for keeping the clean setup of all keys within an account. Similar issue is with function call keys with infinite allowance.

Besides that it would be nice to be able to manage the keys straight away from the table that is an output of the call. Ideally, a user can navigate the keys using arrows and able to delete them with one key press. The flow should allow for multiple key deletions.

At the moment the cleanup of the keys associated with the account requires periodic display of the key list and manual copying of the keys and deleting them one by one.

FroVolod:
Resolves #267

### DIFF
--- a/src/commands/account/delete_key/mod.rs
+++ b/src/commands/account/delete_key/mod.rs
@@ -1,79 +1,36 @@
+mod public_keys_to_delete;
+
 #[derive(Debug, Clone, interactive_clap::InteractiveClap)]
 #[interactive_clap(input_context = crate::GlobalContext)]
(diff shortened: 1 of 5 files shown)


FroVolod:
Screenshot 2023-12-03 at 11 43 29

frol (approved):

PR #272, opened 2023-12-05
feat: Removed the need for an additional network selection step if there is only one network

### LINKED ISSUE #268: Add a setting to fix a network
At the moment, every time when a user is interacting with near cli rs, he needs to specify a network where he's operating. It would be good to have a config to firmly set a network to use.
Even the deletion of the default `testnet` network (which brings us to the list of a single mainnet network) is not removing the need of an additional step of choosing the network, which should be the case.

FroVolod:
Resolves #268


src/commands/account/delete_account/mod.rs
```diff
@@ -152,7 +152,14 @@ impl BeneficiaryAccount {
...
                 .is_none()
                 {
-                    eprintln!("\nHeads up! You will lose remaining NEAR tokens on the account you delete if you specify the account <{account_id}> as the beneficiary as it does not exist.");
+                    let networks: Vec<String> = context
+                        .global_context
+                        .config
+                        .network_connection
+                        .iter()
+                        .map(|(_, network_config)| network_config.network_name.clone())
+                        .collect();
```
frol, line 161: Make it a method on `config`, so you can just call it like:

```
global_context.config.network_names()
```

Use it in every place where you currently copied this code manually.

src/commands/account/delete_key/public_keys_to_delete.rs
```diff
@@ -111,6 +111,21 @@ impl PublicKeyList {
...
+        if access_key_list.is_empty() {
+            let networks: Vec<String> = context
+                .global_context
+                .config
+                .network_connection
+                .iter()
+                .map(|(_, network_config)| network_config.network_name.clone())
+                .collect();
+            return color_eyre::eyre::Result::Err(color_eyre::eyre::eyre!(
+                "Access keys for <{}> on [{}] networks were not found",
```
frol, line 123:
```suggestion
                "Access keys for <{}> were not found on [{}] network(s)",
```

frol (changes requested):

src/common.rs
```diff
@@ -1303,7 +1309,9 @@ fn path_directories() -> Vec<std::path::PathBuf> {
 pub fn get_delegated_validator_list_from_mainnet(
     network_connection: &linked_hash_map::LinkedHashMap<String, crate::config::NetworkConfig>,
 ) -> color_eyre::eyre::Result<std::collections::BTreeSet<near_primitives::types::AccountId>> {
-    let network_config = network_connection.get("mainnet").expect("Internal error!");
+    let network_config = network_connection
+        .get("mainnet")
+        .expect("There is no 'mainnet' network in your configuration.");
```
frol, line 1314: Let's use `wrap_err()?` instead of expect

src/common.rs
```diff
@@ -1794,6 +1802,15 @@ pub fn input_network_name(
     config: &crate::config::Config,
     account_ids: &[near_primitives::types::AccountId],
 ) -> color_eyre::eyre::Result<Option<String>> {
+    if config.network_connection.keys().len() == 1 {
```
frol, line 1805:
```suggestion
    if config.network_connection.len() == 1 {
```

frol (changes requested):

src/common.rs
FroVolod: Done

frol (dismissed):

frol (approved):

PR #293, opened 2024-01-31
feat: Show contract method names

### LINKED ISSUE #70: Show contract method names

FroVolod:
Resolves #70

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,7 @@ url = { version = "2", features = ["serde"] }
 open = "4"
 shell-words = "1"
 cargo-util = "0.1.1"
+wasmer = "4.2.5"
 
 semver = { version = "1.0.4", optional = true }
 self_update = { version = "0.37.0", features = [
--- a/src/commands/contract/inspect_contract/mod.rs
+++ b/src/commands/contract/inspect_contract/mod.rs
@@ -1,0 +1,80 @@
+use color_eyre::eyre::Context;
+
+use crate::common::JsonRpcClientExt;
+
+#[derive(Debug, Clone, interactive_clap::InteractiveClap)]
+#[interactive_clap(input_context = crate::GlobalContext)]
+#[interactive_clap(output_context = ContractContext)]
+pub struct Contract {
+    #[interactive_clap(skip_default_input_arg)]
+    /// What is the contract account ID?
+    account_id: crate::types::account_id::AccountId,
+    #[interactive_clap(named_arg)]
+    /// Select network
+    network_config: crate::network_view_at_block::NetworkViewAtBlockArgs,
+}
+
+impl Contract {
+    pub fn input_account_id(
+        context: &crate::GlobalContext,
+    ) -> color_eyre::eyre::Result<Option<crate::types::account_id::AccountId>> {
+        crate::common::input_non_signer_account_id_from_used_account_list(
+            &context.config.credentials_home_dir,
+            "What is the contract account ID?",
+        )
+    }
+}
+
+#[derive(Clone)]
+pub struct ContractContext(crate::network_view_at_block::ArgsForViewContext);
+
+impl ContractContext {
+    pub fn from_previous_context(
+        previous_context: crate::GlobalContext,
+        scope: &<Contract as interactive_clap::ToInteractiveClapContextScope>::InteractiveClapContextScope,
+    ) -> color_eyre::eyre::Result<Self> {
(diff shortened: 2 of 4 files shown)


Cargo.toml
```diff
@@ -81,6 +82,10 @@ near-token = { version = "0.2.0", features = [
     "interactive-clap",
 ] }
 
+near-abi = "0.4.2"
+zstd = "0.11"
+near-contract-standards = "5.0.0-alpha.2"
```
frol, line 87: This is a fat dependency, let's avoid it. Just implement or copy ContractSourceMetadata here.

FroVolod: Done

src/commands/contract/download_contract_abi/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+        scope: &<DownloadContractAbi as interactive_clap::ToInteractiveClapContextScope>::InteractiveClapContextScope,
+    ) -> color_eyre::eyre::Result<Self> {
+        let on_after_getting_block_reference_callback: crate::network_view_at_block::OnAfterGettingBlockReferenceCallback = std::sync::Arc::new({
+            let account_id = previous_context.account_id.clone();
+            let file_path: std::path::PathBuf = scope.file_path.clone().into();
+
+            move |network_config, block_reference| {
+                if let Ok(contract_abi_response) = tokio::runtime::Runtime::new()
+                    .unwrap()
+                    .block_on(super::inspect_contract::get_contract_abi(&network_config.network_name, &network_config.json_rpc_client(), block_reference, &account_id))
```
frol, line 76: I would make `get_contract_abi` function return the AbiRoot (for the Ok variant), so you don't need to parse it here.

src/commands/contract/inspect_contract/mod.rs
```diff
@@ -0,0 +1,456 @@
...
+                for function in abi_root.body.functions {
+                    let mut table_func = prettytable::Table::new();
+                    table_func.set_format(*prettytable::format::consts::FORMAT_CLEAN);
+                    table_func.add_empty_row();
+
+                    if let near_abi::AbiParameters::Borsh { args: _ } = function.params {
+                        panic!("Borsh is currently unsupported")
+                    }
+                    table_func.add_row(prettytable::row![format!(
+                        "{} (read-write function - {}) {}\n{}",
```
frol, line 220: It must be `read-only function` for "view" functions

src/commands/contract/inspect_contract/mod.rs
```diff
@@ -0,0 +1,456 @@
...
+        &network_config.network_name,
+        &json_rpc_client,
+        &block_reference,
+        account_id,
+    )
+    .await
+    {
+        if let Ok(contract_source_metadata) = contract_source_metadata_response
+            .call_result()?
+            .parse_result_from_json::<self::contract_metadata::ContractSourceMetadata>(
```
frol, line 171: `get_contract_source_metadata` should better return `ContractSourceMetadata` type instead of the response type.

frol (changes requested):
Bugs:

* [ ] Inspecting a contract account without ABI returns nothing

src/commands/contract/download_contract_abi/mod.rs
FroVolod: Done

src/commands/contract/mod.rs
```diff
@@ -26,9 +28,17 @@ pub enum ContractActions {
...
+    ))]
+    /// Get a list of available function names
+    Inspect(self::inspect_contract::Contract),
+    #[strum_discriminants(strum(message = "download-abi    - Download contract ABI"))]
+    /// Download contract ABI
+    DownloadAbi(self::download_contract_abi::Contract),
     #[strum_discriminants(strum(message = "download-wasm   - Download wasm"))]
     /// Download wasm
-    DownloadWasm(self::download_wasm::ContractAccount),
+    DownloadWasm(self::download_wasm::Contract),
```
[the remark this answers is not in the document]
FroVolod: Done

src/commands/contract/inspect_contract/mod.rs
FroVolod: Done

src/commands/contract/inspect_contract/mod.rs
FroVolod: Done

frol (dismissed):

frol (approved):

PR #319, opened 2024-04-13
feat: Support ARM64 Linux

frol:
I tried to follow the suggestion from the cargo-dist thread: https://github.com/axodotdev/cargo-dist/issues/74#issuecomment-2053680080

UPD: No luck:

* libudev is the blocker for musl and cross-compiled aarch64 targets and there is no way to disable ledger support https://github.com/axodotdev/cargo-dist/issues/772
* openssl is another the blocker for musl targets, it needs `aarch64-linux-musl-gcc`

image


frol:
Superseded by #397

PR #325, opened 2024-04-27
feat: improved fetching staking pools

frolvanya:
I've noticed that `view account summary` command was taking too much time. The problems lay in scanning through validators to find if the user’s account id belongs to it. Furthermore, `get_validators_stake` function didn't give all available validators, only +-240 out of 450 were scannable. To fix this, I used fastnear API to retrieve validators in one request, if this process fails, then it will switch to the previous strategy

*before*
image
*after*
image

### DIFF
--- a/src/commands/account/view_account_summary/mod.rs
+++ b/src/commands/account/view_account_summary/mod.rs
@@ -23,6 +23,8 @@ impl ViewAccountSummaryContext {
         previous_context: crate::GlobalContext,
         scope: &<ViewAccountSummary as interactive_clap::ToInteractiveClapContextScope>::InteractiveClapContextScope,
     ) -> color_eyre::eyre::Result<Self> {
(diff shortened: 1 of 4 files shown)


src/commands/account/view_account_summary/mod.rs
```diff
@@ -23,6 +23,8 @@ impl ViewAccountSummaryContext {
         previous_context: crate::GlobalContext,
         scope: &<ViewAccountSummary as interactive_clap::ToInteractiveClapContextScope>::InteractiveClapContextScope,
     ) -> color_eyre::eyre::Result<Self> {
+        let stake_delegators_api = previous_context.config.stake_delegators_api.clone();
```
frolvanya, line 26: I'm not sure if this is good practice

frol (approved):

PR #333, opened 2024-05-05
feat: displaying transaction fee 

### LINKED ISSUE #320: Display the transaction fee in NEAR and approximate $ after transaction is executed
So after any transaction being executed, we currently display logs, status, and link to Explorer, and I suggest we add a summary of the transaction fees in NEAR and USD equivalent.

image

Here is the function you need to apply the changes: https://github.com/near/near-cli-rs/blob/a1bde9a58dbba0fdaa7813d02752488cd9ec761e/src/common.rs#L1102-L1107

frolvanya:
Displaying a simple transaction fee after transaction execution
This PR closes #320 

image

### DIFF
--- a/src/commands/config/add_connection/mod.rs
+++ b/src/commands/config/add_connection/mod.rs
@@ -302,7 +302,7 @@ impl AddNetworkConnection {
         if let ConfirmOptions::Yes = select_choose_input {
             let coingecko_api: crate::types::url::Url =
                 CustomType::new("What is the coingecko API url?")
-                    .with_starting_input("https://api.coingecko.com")
+                    .with_starting_input("https://api.coingecko.com/")
                     .prompt()?;
             Ok(Some(coingecko_api))
         } else {
--- a/src/commands/config/add_connection/mod.rs
+++ b/src/commands/config/add_connection/mod.rs
@@ -36,10 +36,13 @@ pub struct AddNetworkConnection {
     meta_transaction_relayer_url: Option<crate::types::url::Url>,
     #[interactive_clap(long)]
     #[interactive_clap(skip_default_input_arg)]
-    fastnear_url: Option<String>,
+    fastnear_url: Option<crate::types::url::Url>,
     #[interactive_clap(long)]
     #[interactive_clap(skip_default_input_arg)]
     staking_pools_factory_account_id: Option<crate::types::account_id::AccountId>,
+    #[interactive_clap(long)]
+    #[interactive_clap(skip_default_input_arg)]
+    coingecko_url: Option<crate::types::url::Url>,
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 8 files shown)


src/common.rs
```diff
@@ -1123,6 +1123,30 @@ pub fn handler_invalid_tx_error(
...
+            return Ok(Some(price));
+        }
+    }
+
+    Ok(None)
+}
+
+fn convert_near_to_usd(near: crate::types::near_token::NearToken) -> Option<f64> {
+    let near_price = get_near_price().ok().flatten();
+    near_price.map(|price| near.0.as_yoctonear() as f64 * price / 10_f64.powf(24_f64))
```
frolvanya, line 1147: This is a bad way of converting NEAR to USD. I'll need to fix it later

frolvanya:
PR seems to be complete, however, there's strange bug related to tracing from "send" module. It prints "Sending transaction ..." twice
https://github.com/near/near-cli-rs/blob/8904db755435ee8c6ccb5bebf47fa55f88ada018/src/transaction_signature_options/send/mod.rs#L13
image

src/common.rs
```diff
@@ -1123,6 +1123,35 @@ pub fn handler_invalid_tx_error(
...
+        let body = response.text()?;
+        let parsed_body: serde_json::Value = serde_json::from_str(&body)?;
+
+        let price = parsed_body["near"]["usd"].as_f64();
+
+        if let Some(price) = price {
+            return Ok(Some(price));
+        }
+
+        std::thread::sleep(std::time::Duration::from_millis(100));
```
frol, line 1145: This implementation is not going to retry. If request will fail, it will return Err on `?`. Fix the implementation.

src/common.rs
```diff
@@ -1123,6 +1123,35 @@ pub fn handler_invalid_tx_error(
...
+        api.join("api/v3/simple/price?ids=near&vs_currencies=usd")?
+    } else {
+        return Ok(None);
+    };
+
+    for _ in 0..10 {
+        let response = reqwest::blocking::get(url.clone())?;
+
+        let body = response.text()?;
+        let parsed_body: serde_json::Value = serde_json::from_str(&body)?;
```
frol, line 1137: Instead of getting `.text()` and then manually parsing it, use `.json()` method.

```rust
if let Ok(parsed_body) = response.json::<serde_json::Value>()
```

src/common.rs
```diff
@@ -1142,14 +1171,41 @@ pub fn print_transaction_status(
...
             eprintln!("Logs [{}]:", receipt.outcome.executor_id);
             eprintln!("  {}", receipt.outcome.logs.join("\n  "));
         };
     }
+
+    let approximate_usd = if let Some(price) = price {
+        format!("{:.8}", convert_near_to_usd(total_tokens_burnt, price))
+    } else {
+        "N/A".to_string()
+    };
```
frol, line 1197: "(approximate $N/A USD)" is not helpful, so let's remove the whole block from formatting in that case.

src/commands/config/add_connection/mod.rs
```diff
@@ -272,4 +282,29 @@ impl AddNetworkConnection {
...
+            No,
+        }
+        let select_choose_input = Select::new(
+            "Do you want to enter the coingecko API url?",
+            vec![ConfirmOptions::Yes, ConfirmOptions::No],
+        )
+        .prompt()?;
+        if let ConfirmOptions::Yes = select_choose_input {
+            let coingecko_api: crate::types::url::Url =
+                CustomType::new("What is the coingecko API url?").prompt()?;
```
frol, line 304: Use [`.with_starting_input()`](https://docs.rs/inquire/latest/inquire/struct.CustomType.html#method.with_starting_input) to prepopulate the input with the default coingecko API URL.

frol (changes requested):

frol (dismissed):

frol (approved):

frol (approved):

Issue #299, opened 2024-02-10
Implement keychain support in WSL (Linux on Windows)


frol:
It seems that near.cli.rs cannot use system keychain to store and read keys on WSL (Linux subsystem on Windows): https://github.com/near/near-cli-rs/issues/287#issuecomment-1935586950

It needs further investigation if there is a way to get system keychain working in WSL and if there is a need to change anything in near.cli.rs for that.

Meanwhile, if you use WSL, use legacy keychain to manage the keys.

AlexKushnir1:
I can work on this!

PR #347, opened 2024-05-27
Give instructions that help to resolve a problem with missing keychain

AlexKushnir1:
I found out that certain headless linux systems don't implement keyring. For example on WSL we can`t use keychain because it is not seted up by default. User need to install gnome-keyring with pip3, start D-bust session and unlock keychain. Close https://github.com/near/near-cli-rs/issues/299

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -1267,7 +1267,7 @@ pub fn save_access_key_to_keychain(
     keyring::Entry::new(&service_name, &format!("{}:{}", account_id, public_key_str))
         .wrap_err("Failed to open keychain")?
         .set_password(key_pair_properties_buf)
-        .wrap_err("Failed to save password to keychain")?;
+        .wrap_err("Failed to save password to keychain. You may have to install the secure keychain by the following instructions: https://github.com/jaraco/keyring#using-keyring-on-headless-linux-systems")?;
 
     Ok("The data for the access key is saved in the keychain".to_string())
 }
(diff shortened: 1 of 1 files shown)


src/common.rs
```diff
@@ -1267,7 +1267,7 @@ pub fn save_access_key_to_keychain(
     keyring::Entry::new(&service_name, &format!("{}:{}", account_id, public_key_str))
         .wrap_err("Failed to open keychain")?
         .set_password(key_pair_properties_buf)
-        .wrap_err("Failed to save password to keychain")?;
+        .wrap_err("Failed to save password to keychain. You may have to install the secure keychain by the following instructions: https://github.com/jaraco/keyring#using-keyring-on-headless-linux-systems")?;
```
volovyks, line 1270:
```suggestion
        .wrap_err("Failed to save password to keychain. You may need to install secure keychain package by following  this instructions: https://github.com/jaraco/keyring#using-keyring-on-headless-linux-systems")?;
```

volovyks (approved):

src/common.rs
```diff
@@ -1267,7 +1267,7 @@ pub fn save_access_key_to_keychain(
     keyring::Entry::new(&service_name, &format!("{}:{}", account_id, public_key_str))
         .wrap_err("Failed to open keychain")?
         .set_password(key_pair_properties_buf)
-        .wrap_err("Failed to save password to keychain")?;
+        .wrap_err("Failed to save password to keychain. You may need to install secure keychain package by following  this instructions: https://github.com/jaraco/keyring#using-keyring-on-headless-linux-systems")?;
```
frolvanya, line 1270: - Is it okay that we have a double-space between words (`following  this`)?
- Last sentence isn't grammatically correct, it should be `these instructions` or `this instruction` (I guess first one is better)
- `the secure keychain` looked correct in a previous version. Why an article was removed in the last commit?

frol (approved):
@alexkushnir1 Thank you for taking a look there!

PR #353, opened 2024-06-20
Proof of concept implementation of the Olaf protocol

Fiono11:
### What

This is a proof of concept implementation of the [Olaf](https://eprint.iacr.org/2023/899) protocol made with this [grant](https://gov.near.org/t/proposal-frost-implementation-compatible-with-near/38260). Olaf is composed of the SimplPedPoP protocol to generate a threshold account, and the FROST protocol to sign from a threshold account.

### Why

It is much more efficient than the current multisignature protocol used by Near, since the signing is done off chain instead of on chain. The advantages of this are:

- Only one signature needs to be verified, which is independent of the number of signers. This is more scalable (up to 1000 signers) and cheaper.
- More private, since the network does not know which individual accounts partially signed the threshold signature.

### Limitations

Right now communication between parties needs to be done out-of-band, since networking is not implemented yet. In this draft PR there is a tutorial with an example of a 2-of-2 payment that can be found in `docs`.

### Future Work

Let me know if there is interest in continuing this project. Next steps would be:

- Implement networking with the WebRTC protocol.
- Integrate it with a wallet from the Near ecosystem.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2540,7 +2540,6 @@ dependencies = [
  "cargo-util",
  "clap",
  "color-eyre",
- "curve25519-dalek 4.1.3",
  "derive_more",
  "dirs",
  "easy-ext 1.0.2",
@@ -2587,7 +2586,6 @@ dependencies = [
  "tracing-subscriber",
  "url",
  "wasmparser",
- "zeroize",
  "zstd",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "actix"
-version = "0.13.3"
+version = "0.13.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fb72882332b6d6282f428b77ba0358cb2687e61a6f6df6a6d3871e8a177c2d4f"
+checksum = "de7fa236829ba0841304542f7614c42b80fca007455315c45c785ccfa873a85b"
 dependencies = [
  "actix-macros",
  "actix-rt",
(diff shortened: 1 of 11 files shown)


frol:
@Fiono11 Wow, that is intense! I think it is a good candidate to actually be a separate NEAR CLI extension like https://github.com/near-cli-rs/near-validator-cli-rs

It is fine to keep it in the PR if you like it, but I want to be clear upfront that we will need to answer the question of why would it be a good addition to all the near-cli-rs users. So far it seems that NEAT CLI extension is the better direction to explore; note that `near XXX` commands, where XXX is an unknown command, are translated into `near-XXX ...` calls similar to how `cargo` does it (we have [cargo-near](https://github.com/near/cargo-near), [bos-cli-rs](https://github.com/bos-cli-rs/bos-cli-rs) that use this same "framework" we have in near-cli-rs, so you may take inspiration from there).

Fiono11:
> @Fiono11 Wow, that is intense! I think it is a good candidate to actually be a separate NEAR CLI extension like https://github.com/near-cli-rs/near-validator-cli-rs
> 
> It is fine to keep it in the PR if you like it, but I want to be clear upfront that we will need to answer the question of why would it be a good addition to all the near-cli-rs users. So far it seems that NEAT CLI extension is the better direction to explore; note that `near XXX` commands, where XXX is an unknown command, are translated into `near-XXX ...` calls similar to how `cargo` does it (we have [cargo-near](https://github.com/near/cargo-near), [bos-cli-rs](https://github.com/bos-cli-rs/bos-cli-rs) that use this same "framework" we have in near-cli-rs, so you may take inspiration from there).

Thanks! That's what I wanted to know: what is the right direction to go. And it makes sense that an extension is the right approach, since the protocol is done off chain. Will have a look at it.

PR #358, opened 2024-06-26
feat: Added load waiting indicators for "transaction" group commands 

FroVolod:


src/commands/transaction/construct_transaction/add_action_3/add_action/stake/mod.rs
```diff
@@ -2,7 +2,9 @@
 #[interactive_clap(input_context = super::super::super::ConstructTransactionContext)]
 #[interactive_clap(output_context = StakeActionContext)]
 pub struct StakeAction {
+    /// Enter the amount to stake: (example: 10000NEAR)
     stake_amount: crate::types::near_token::NearToken,
+    /// Validator key which will be used to sign transactions on behalf of signer_id:
```
frol, line 7:
```suggestion
    /// Enter the public key of the validator key pair used on your NEAR node (see validator_key.json):
```

frol (changes requested):

frol (approved):

PR #380, opened 2024-08-07
chore: Disable env section of color_eyre report

FroVolod:
[inspired by this example: https://github.com/near/cargo-near/pull/196](https://github.com/near/cargo-near/pull/196)

@race-of-sloths


frol (approved):
@race-of-sloths score 1

PR #381, opened 2024-08-09
chore: updated near-* to 0.24.0, bumped up rust version

akorchyn:
Resolves the issue that the `near tokens user view-near-balance` fails to run
After PR: 
image
Before PR:
image

Resolves: #384 

CC @race-of-sloths


frol:
Please, update interactive-clap as well

frol (approved):
@race-of-sloths score 3

PR #414, opened 2024-11-13
chore!: updates near-* dependencies to 0.27 release

akorchyn:
@race-of-sloths


dj8yfo:
@race-of-sloths score 3

PR #420, opened 2024-11-21
fix: deprecate `wallet.near.org` & `explorer.near.org`

### LINKED ISSUE #418: [BUG] Transaction link uses deprecated `explorer.near.org` address
### Background

Transaction links provided in the console summary use the deprecated `explorer.near.org` address and should use `nearblocks.io` or similar. 

`Example:`

```bash
Gas burned: 17.0 Tgas
Transaction fee: 0.001608566867888 NEAR
Transaction ID: 3TBv37NiKQuao9Rou5dHxh3kezEpfrMLCNN48zhNRqrx
To see the transaction in the transaction explorer, please open this url in your browser:
https://explorer.testnet.near.org/transactions/3TBv37NiKQuao9Rou5dHxh3kezEpfrMLCNN48zhNRqrx
```

### User Story

As a developer I would expect to be directed to a working site :) 

### Acceptance Criteria

- [ ] update both `mainnet` and `testnet` explorer links to use nearblocks.io or similar

### Resources & Additional Notes

_No response_

### Priority

🟡 P2 : Medium

thisisjoshford:
Fixes #418 

As wallet.near.org & explorer.near.org are no longer active, this PR updates these urls to active projects (app.mynearwallet.com & nearblocks.io)


frol:
Please, keep explorer.near.org - it is a selector of all the available Explorers, instead of hardcoding one specific Explorer there.

thisisjoshford:
> Consider keeping explorer.near.org - it is a selector of all the available Explorers, instead of hardcoding one specific Explorer there.

Are you suggesting we send them to the explorer landing site and then tell them to choose one and search for this tx? or should we have both a direct link to tx and the general landing page for visibility on other options they have?

I particularly like having a smooth UX of a one click to tx result, but I see your point in not forcing one solution over another. If we wanted to be super slick we could have the user choose which explorer they would want to use and then it saves this configuration for future CLI transactions :)

gagdiez:
@frol just came across this PR, and I want to say that I agree with @thisisjoshford, it is better to be sent directly to a explorer to reduce friction in the UX

frol:
I am posting here for visibility: We discussed it over the call with @thisisjoshford and he did not realize that explorer.near.org handles custom paths, so user will be redirected to the appropriate page, and not the home page of the selected Explorer.

image

Also, we agreed that it is great to have this Selector to let users discover alternative Explorers and yet we may introduce a checkbox "save my choice and redirect me automatically in the future" on the Selector page to further streamline the usage of the Selector.

No actions needed on NEAR CLI side.

PR #459, opened 2025-03-19
feat: Added exit codes.

FroVolod:
Standard [sysexits(3)](https://man7.org/linux/man-pages/man3/sysexits.h.3head.html) codes have been added for exit codes.

Screenshot 2025-03-19 at 17 08 31

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2116,6 +2116,7 @@ dependencies = [
  "smart-default",
  "strum",
  "strum_macros",
+ "sysexits",
  "textwrap",
  "thiserror 2.0.6",
  "tokio",
@@ -3680,6 +3681,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "sysexits"
+version = "0.8.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3a70e00b0ea6c3e7154dfc48fee004bf61bdc154fd99b7e17c05318503b9660b"
+
+[[package]]
 name = "system-configuration"
 version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


dj8yfo (commented):
I've looked though this a bit, 
1. this depends on using some low-level machinery with vtables from 
`eyre` : https://github.com/eyre-rs/eyre/blob/master/eyre/src/error.rs#L646-L665 and https://github.com/eyre-rs/eyre/blob/master/eyre/src/error.rs#L717-L733, which is extremely hard to understand thoroughly (at least to me). 
2. It's hard to tell whether the error codes will bubble up in https://github.com/near/near-cli-rs/blob/7931b8690422e17099de14be8a7861bce150147e/src/main.rs#L274 if there're a few  layers of `wrap_` methods used, but `context_chain_downcast` implementation suggest it will, as it calls `(source.vtable().object_downcast)` down the chain. Thus it requires testing of a multitude of different cases by inducing errors to check whether the added error codes are really returned by the program (it's not obvious just from the fact that the code compiles due to the mentioned above trouble with understanding vtables magic). 
3. https://docs.rs/anyhow/1.0.98/anyhow/struct.Error.html#method.downcast_ref from `anyhow` (not from `eyre`) is already used in https://github.com/near/near-cli-rs/blob/8f7d78a37def692a9b598947fd2ee3af682abc0d/src/common.rs#L1676 prior to this pr. 
4. This pr makes the code of the repo more long/complex, and i'd be happier to see `main()` function split into smaller functions or splitting 3k `common.rs` module into submodules. 
5. This is not the smallest chunk to review and check nothing was broken/new bug introduced. This pr has already seen a few merges, so an error is possible solely due to a merge artifact, which has happened before. 

Finally, i think it's unlikely someone will use exit codes to dispatch behavior in complex scenarios in shell scripts, and if someone does need it, he/she can use a script in rust, programmable with `near-api` and achieve more predictable results with it.

PR #462, opened 2025-03-27
fix: Fixed returning the result of a function call

FroVolod:
Screenshot 2025-03-27 at 21 08 00

### DIFF
--- a/src/commands/contract/call_function/as_read_only/mod.rs
+++ b/src/commands/contract/call_function/as_read_only/mod.rs
@@ -1,4 +1,5 @@
 use color_eyre::eyre::Context;
+use std::io::Write;
 
 use crate::common::CallResultExt;
 use crate::common::JsonRpcClientExt;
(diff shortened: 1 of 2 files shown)


src/commands/contract/call_function/as_read_only/mod.rs
```diff
@@ -137,29 +138,33 @@ fn call_view_function(
                 function_name, account_id, network_config.network_name
             )
         })?;
-    call_result.print_logs();
-    let mut info_str = String::new();
-    if call_result.result.is_empty() {
-        info_str.push_str("\nEmpty result");
+
+    if let crate::Verbosity::Quiet = verbosity {
```
akorchyn, line 142: Move that to the 157 line. It can be if else instead of two checks...

src/common.rs
```diff
@@ -1452,6 +1451,28 @@ pub fn print_transaction_status(
             } else {
                 "The returned value is not printable (binary data)".to_string()
             };
+            if let crate::Verbosity::Interactive = verbosity {
+                for action in &transaction_info.transaction.actions {
+                    if let near_primitives::views::ActionView::FunctionCall {
```
akorchyn, line 1456: if let near_primitives::views::ActionView::FunctionCall { .. }

akorchyn (approved):
some nits

src/commands/contract/call_function/as_read_only/mod.rs
FroVolod: I disagree with you. Since I'm considering three verbosity options (Quiet, Interactive, TeachMe), it's better not to perform unnecessary operations.

src/common.rs
FroVolod: Done

src/commands/contract/call_function/as_read_only/mod.rs
akorchyn: To have this printing in one place rather than at the top of functions, and at the end of the functions. It's harder to read the code personally

FroVolod: Done

PR #478, opened 2025-04-29
fix: Fixed canceling the CLI operation

### LINKED ISSUE #473: Ctrl-C doesn't exit CLI if it has connectivity issue checking whether an account exists
Image

FroVolod:
Resolves #473


frol (approved):
@race-of-sloths score 5

PR #487, opened 2025-05-22
chore: fixed shell configuration for command history

### LINKED ISSUE #479: Shell integration doesn't escape quotes in json strings
When I do `near contract call-function as-transaction pre.intear.near ft_transfer_call json-args '{"receiver_id":"intear.pool.near","amount":"500000000000000000000000000000","msg":"{\"name\":\"Pre-TEAR\",\"start_date\":\"1746255906617000000\",\"end_date\":\"1748725200000000000\",\"farm_id\":null}"}' prepaid-gas '100.0 Tgas' attached-deposit '1 yoctoNEAR' sign-as intear.near network-config mainnet sign-with-keychain send`
The command that gets appended to shell history (https://github.com/near/near-cli-rs/blob/main/docs/SHELL_HISTORY_INTEGRATION.md) is `near contract call-function as-transaction pre.intear.near ft_transfer_call json-args '{"receiver_id":"intear.pool.near","amount":"500000000000000000000000000000","msg":"{"name":"Pre-TEAR","start_date":"1746255906617000000","end_date":"1748725200000000000","farm_id":null}"}' prepaid-gas '100.0 Tgas' attached-deposit '1 yoctoNEAR' sign-as intear.near network-config mainnet sign-with-keychain send`. The JSON object inside `msg` string is not escaped, so when I run this command from history, it doesn't parse as JSON.

Didn't test with other shells.

vsavchyn-dev:
This pull requests closes #479 

Resolves the problem with zsh and Fish shell scripts, not being able to write escaped characters properly into history

### DIFF
--- a/docs/SHELL_HISTORY_INTEGRATION.md
+++ b/docs/SHELL_HISTORY_INTEGRATION.md
@@ -40,7 +40,7 @@ function near() {
         final_command=$(<"$tmp_file")
 
         if [[ -n "$final_command" ]]; then
-            print -s -- "$final_command"
+            print -sr -- "$final_command"
         fi
 
         rm "$tmp_file"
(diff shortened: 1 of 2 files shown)


dj8yfo:
@vsavchyn-dev can you confirm that the printed command at the end of near-cli execution for problem case in #479 is correct, and it's only shell integration instruction that is slightly off? Otherwise, the fix is in incorrect place, as near-cli has to be fixed then. 
@frolvanya can you review this and confirm the update is fine? As you added support for integration initially.

vsavchyn-dev:
@dj8yfo the problem is only with shell integration script. Problem present in zsh and fish, as it writes to shell history using echo conventions, converting `\'` to `'`, which messes up how it reads from history later. That's why in #479 initial command wasn't same as one present in history.

dj8yfo (approved):
@race-of-sloths score 13

vsavchyn-dev:
@dj8yfo idk how to fix it, so I silenced it for now...
will create issue soon.

vsavchyn-dev:
> @vsavchyn-dev can you create another pr with fixing clippy on 1.87?

@dj8yfo done in pr https://github.com/near/near-cli-rs/pull/488

PR #490, opened 2025-05-27
refactor: introduced `BoxedJsonRpcResult` type

frolvanya:
This is done to reduce code repetition a bit

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -16,7 +16,10 @@ use tracing_indicatif::suspend_tracing_indicatif;
 use near_primitives::{hash::CryptoHash, types::BlockReference, views::AccessKeyPermissionView};
 
 pub type CliResult = color_eyre::eyre::Result<()>;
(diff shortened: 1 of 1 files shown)


src/common.rs
```diff
@@ -16,6 +16,7 @@ use tracing_indicatif::suspend_tracing_indicatif;
 use near_primitives::{hash::CryptoHash, types::BlockReference, views::AccessKeyPermissionView};
 
 pub type CliResult = color_eyre::eyre::Result<()>;
+pub type JsonRpcResult<T, E> = Result<T, Box<near_jsonrpc_client::errors::JsonRpcError<E>>>;
```
dj8yfo, line 19: i'd rename to `BoxedJsonRpcResult` and add comment about clippy's lint, which caused the creation of this alias

frolvanya: Makes sense. Fixed in https://github.com/near/near-cli-rs/pull/490/commits/8d78d29b187e24f950dc83386c6a7fa815cd02ad

dj8yfo (approved):
@race-of-sloths score 8

PR #492, opened 2025-06-12
feat: Add support for submitting general DAO proposals

### LINKED ISSUE #197: feat: Add support for submitting DAO proposals
Currently, commands like `near tokens xxx.near send-near` prepare a transaction with native actions such as `TRANSFER` or `FUNCTION_CALL` to FT contract. I tried to integrate DAO commands right into the existing commands, but with DAO we have two different accounts (DAO contract account id, and signer account id) in addition to a receiver account id (if applicable).

**IGNORE THE TEXT BELOW (I keep it only for historical reasons, see my next comment)**

-------------------------------------

Here is the current command to send NEAR:

```
near tokens sender.near send-near receiver.near 10N network-config mainnet
```

If I just try to "send tokens from DAO", there is no way to specify who will be signing the transaction:

```
near tokens dao.near send-near receiver.near 10N network-config mainnet
```

So my natural first attempt was:

```
near dao devdao.near tokens send-near receiver.near 10N sign-as signer.near network-config mainnet
```

Though, it might mislead users that this command will actually send the tokens while it is only intended to submit a new proposal to the DAO with a TRANSFER action, which will only be executed when the proposal is confirmed. So second thought is to introduce `submit-proposal` and `confirm-proposal` commands:

```
near dao devdao.near submit-proposal send-near receiver.near 10n sign-as signer.near network-config mainnet
```

The problem I hate to admit is that we won't be able to reuse existing commands, so `send-near`, `send-ft`, `send-nft`, `function-call as-transaction`, `deploy`, `account manage-storage-deposit deposit / withdraw` commands will need to be re-implemented, and it is even more painful if we take extensions into account, e.g. `bos` CLI components deployment, permissions and prepaid storage management would need to be re-implemented for DAO as well.

At least DAO proposal confirmation seems to be quite straightforward:

```
near dao devdao.near confirm-proposal 123 gas-attached 200TGAS sign-as signer.near network-config mainnet
```

I would love to spend more time thinking about how to avoid code duplication, so this issue is just a placeholder for now.

vsavchyn-dev:
This pull request closes #197, adding support for submitting transactions of type `Action::FunctionCall` and `Action::Transfer` as `FunctionCall` and `Transfer` DAO proposal kinds respectfully.

Therefore, by writing:
```
near \
  tokens devhub.sputnik-dao.near \
  send-near receiver.near 10N \
  network-config mainnet \
  submit-dao-proposal dao-member.near \
  'dao description' \
  prepaid-gas '100.0 Tgas' \
  attached-deposit 1N \
  sign-with-keychain \
  send
```
You can construct `Transfer` dao proposal:
```
signer_id:    dao-member.near
receiver_id:  devhub.sputnik-dao.near
actions:
       -- function call:      
                       method name:  add_proposal
                       args:         {
                                       "proposal": {
                                         "description": "dao description",
                                         "kind": {
                                           "Transfer": {
                                             "amount": "10000000000000000000000000",
                                             "msg": null,
                                             "receiver_id": "receiver.near",
                                             "token_id": ""
                                           }
                                         }
                                       }
                                     }
                       gas:          100.0 Tgas
                       deposit:      1 NEAR
```

### DAO Proposal Functionality:
* Added a new module `submit_dao_proposal` in `src/transaction_signature_options/mod.rs` to support converting transactions into DAO proposals. This includes defining the `DaoProposal` structure and its associated arguments, contexts, and signing options.
* Implemented `ProposalKind` enum in `src/transaction_signature_options/submit_dao_proposal/dao_kind_arguments.rs` to handle different types of DAO proposal arguments, such as `Transfer` and `FunctionCall`.

### Argument Decoding:
* Added a `deep_decode_args` function in `src/common.rs` for `print_unsigned_transaction` function to recursively decode base64-encoded arguments in JSON objects. This ensures nested arguments are properly decoded, so person who submits things can see all sub-arguments in human-readable text, and not in base64.

### Dependency Updates:
* Added new dependencies `base64` and `serde_with` to `Cargo.toml` for handling base64 encoding/decoding and advanced serialization/deserialization (it was already included to other dependencies).

### Utility Enhancements:
* Added a `from_near` method to the `NearToken` type in `src/types/near_token.rs`.

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -4,7 +4,6 @@ use std::fs::OpenOptions;
 use std::io::Write;
 use std::str::FromStr;
 
-use base64::Engine;
 use color_eyre::eyre::{ContextCompat, WrapErr};
 use color_eyre::owo_colors::OwoColorize;
 use futures::{StreamExt, TryStreamExt};
@@ -715,35 +714,6 @@ pub fn print_full_unsigned_transaction(
     info_str
 }
 
-pub fn deep_decode_args(v: &mut serde_json::Value) {
-    match v {
-        serde_json::Value::Object(map) => {
-            for (key, val) in map.iter_mut() {
-                if key == "args" {
-                    if let serde_json::Value::String(s) = val {
-                        if let Ok(bytes) = base64::engine::general_purpose::STANDARD.decode(s) {
-                            if let Ok(mut nested) =
-                                serde_json::from_slice::<serde_json::Value>(&bytes)
-                            {
-                                deep_decode_args(&mut nested);
-                                *val = nested;
-                                continue;
-                            }
-                        }
-                    }
-                }
-                deep_decode_args(val);
-            }
-        }
-        serde_json::Value::Array(arr) => {
-            for elem in arr {
-                deep_decode_args(elem);
-            }
-        }
-        _ => {}
-    }
-}
-
 pub fn print_unsigned_transaction(
     transaction: &crate::commands::PrepopulatedTransaction,
 ) -> String {
@@ -794,9 +764,8 @@ pub fn print_unsigned_transaction(
                     "",
                     "args:",
                     match serde_json::from_slice::<serde_json::Value>(&function_call_action.args) {
-                        Ok(mut parsed) => {
-                            deep_decode_args(&mut parsed);
-                            serde_json::to_string_pretty(&parsed)
+                        Ok(parsed_args) => {
+                            serde_json::to_string_pretty(&parsed_args)
                                 .unwrap_or_else(|_| "".to_string())
                                 .replace('\n', "\n                                 ")
                         }
--- a/src/common.rs
+++ b/src/common.rs
@@ -4,6 +4,7 @@ use std::fs::OpenOptions;
 use std::io::Write;
 use std::str::FromStr;
 
+use base64::Engine;
 use color_eyre::eyre::{ContextCompat, WrapErr};
 use color_eyre::owo_colors::OwoColorize;
 use futures::{StreamExt, TryStreamExt};
@@ -714,6 +715,35 @@ pub fn print_full_unsigned_transaction(
     info_str
 }
 
+pub fn deep_decode_args(v: &mut serde_json::Value) {
+    match v {
+        serde_json::Value::Object(map) => {
+            for (key, val) in map.iter_mut() {
+                if key == "args" {
+                    if let serde_json::Value::String(s) = val {
+                        if let Ok(bytes) = base64::engine::general_purpose::STANDARD.decode(s) {
+                            if let Ok(mut nested) =
+                                serde_json::from_slice::<serde_json::Value>(&bytes)
+                            {
+                                deep_decode_args(&mut nested);
+                                *val = nested;
+                                continue;
+                            }
+                        }
+                    }
+                }
+                deep_decode_args(val);
+            }
+        }
+        serde_json::Value::Array(arr) => {
+            for elem in arr {
+                deep_decode_args(elem);
+            }
+        }
+        _ => {}
+    }
+}
+
 pub fn print_unsigned_transaction(
     transaction: &crate::commands::PrepopulatedTransaction,
 ) -> String {
@@ -764,14 +794,15 @@ pub fn print_unsigned_transaction(
                     "",
                     "args:",
                     match serde_json::from_slice::<serde_json::Value>(&function_call_action.args) {
-                        Ok(parsed_args) => {
-                            serde_json::to_string_pretty(&parsed_args)
+                        Ok(mut parsed) => {
+                            deep_decode_args(&mut parsed);
+                            serde_json::to_string_pretty(&parsed)
                                 .unwrap_or_else(|_| "".to_string())
                                 .replace('\n', "\n                                 ")
                         }
                         Err(_) => {
-                            if let Ok(args) = String::from_utf8(function_call_action.args.clone()) {
-                                args
+                            if let Ok(s) = String::from_utf8(function_call_action.args.clone()) {
+                                s
                             } else {
                                 format!(
                                     "<non-printable data ({})>",
--- a/src/common.rs
+++ b/src/common.rs
@@ -801,8 +801,8 @@ pub fn print_unsigned_transaction(
                                 .replace('\n', "\n                                 ")
                         }
                         Err(_) => {
-                            if let Ok(s) = String::from_utf8(function_call_action.args.clone()) {
-                                s
+                            if let Ok(args) = String::from_utf8(function_call_action.args.clone()) {
+                                args
                             } else {
                                 format!(
                                     "<non-printable data ({})>",
--- a/src/transaction_signature_options/submit_dao_proposal/dao_kind_arguments.rs
+++ b/src/transaction_signature_options/submit_dao_proposal/dao_kind_arguments.rs
@@ -1,3 +1,4 @@
+use color_eyre::eyre::eyre;
 use near_primitives::action::Action;
 use serde::{Deserialize, Serialize};
 use serde_with::{base64::Base64, serde_as, DisplayFromStr};
@@ -9,7 +10,7 @@ pub struct ActionCall {
     #[serde_as(as = "Base64")]
     args: Vec<u8>,
     deposit: crate::types::near_token::NearToken,
-    // NOTE: We cannot use `crate::common::NearGas`, as sputnikdao uses `U128`
+    // NOTE: We cannot use `crate::common::NearGas`, as sputnikdao uses `U64`
     // https://github.com/near-daos/sputnik-dao-contract/blob/278adc713e795f95a6da4d3007c7e03e8120f153/sputnikdao2/src/proposals.rs#L42
     #[serde_as(as = "DisplayFromStr")]
     gas: u64,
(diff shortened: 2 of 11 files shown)


src/transaction_signature_options/submit_dao_proposal/dao_kind_arguments.rs
```diff
@@ -0,0 +1,128 @@
...
+                Action::Transfer(_) => {
+                    if parsed_actions.is_empty() {
+                        parsed_actions.push(action.clone());
+                    } else if let Action::Transfer(_) = parsed_actions.last().unwrap() {
+                        return Err(eyre!("Batch transfers are not supported for DAO proposals"));
+                    } else {
+                        return Err(eyre!(
+                            "Mixed action types are not supported for DAO proposals"
+                        ));
+                    }
```
frol: Each action can be converted into an individual `add_proposal` function call. One transaction can have multiple function call actions.

src/transaction_signature_options/submit_dao_proposal/mod.rs
```diff
@@ -0,0 +1,272 @@
...
+        })
+    }
+}
+
+impl PrepaidGas {
+    pub fn input_gas(
+        _context: &DaoProposalArgumentsContext,
+    ) -> color_eyre::eyre::Result<Option<crate::common::NearGas>> {
+        Ok(Some(
+            CustomType::new("Enter gas for addding DAO proposal:")
```
frol:
```suggestion
            CustomType::new("What is the gas limit for adding DAO proposal (if unsure, keep 10 TGas)?")
```

src/transaction_signature_options/submit_dao_proposal/mod.rs
```diff
@@ -0,0 +1,272 @@
...
+                .with_starting_input("100 TeraGas")
```
frol: Let's reduce the default to 10 TeraGas. I checked and could not really find any `add_proposal` that is more than 3 TeraGas, the majority is in the range of 2.4-2.6 TGas.

```suggestion
                .with_starting_input("10 Tgas")
```

src/transaction_signature_options/submit_dao_proposal/mod.rs
```diff
@@ -0,0 +1,272 @@
...
+
+impl Deposit {
+    pub fn input_deposit(
+        _context: &PrepaidGasContext,
+    ) -> color_eyre::eyre::Result<Option<crate::types::near_token::NearToken>> {
+        let one_near = crate::types::near_token::NearToken::from_near(1);
+
+        Ok(Some(
+            CustomType::new("Enter deposit for adding DAO proposal:")
+                .with_starting_input(&one_near.to_string())
```
frol: Let's use 0 NEAR as default and do not enforce the minimum amount as 0 NEAR is a totally valid value.

```suggestion
        Ok(Some(
            CustomType::new("Enter deposit for adding DAO proposal:")
                .with_starting_input("0 NEAR")
```

src/common.rs
```diff
@@ -764,8 +794,9 @@ pub fn print_unsigned_transaction(
                     "",
                     "args:",
                     match serde_json::from_slice::<serde_json::Value>(&function_call_action.args) {
-                        Ok(parsed_args) => {
-                            serde_json::to_string_pretty(&parsed_args)
+                        Ok(mut parsed) => {
+                            deep_decode_args(&mut parsed);
```
frol: Let's not do this as this will be confusing when someone will sign the transaction via Ledger device and see that the actual arguments are different. I agree that base64 is not easy to review, but we have the actual args printed a few lines above (when user selects "sing-*" or "submit-as-dao-proposal"), so we should be fine.

frol (changes requested):
@vsavchyn-dev Niiice! This looks really good and already works mostly fine. I would like to only ask to address the minor review requests and we are good to go!

src/common.rs
frolvanya: Why is this confusing for Ledger user? I'd actually prefer to see full arguments when adding/verifying the proposal. It would be more confusing to have different args between logs inside one execution

src/transaction_signature_options/submit_dao_proposal/mod.rs
frolvanya: `add_proposal` requires 1 NEAR by default, so IMO it'd make more sense to leave it like this

src/common.rs
frol: If I would see `{"args": "e30="}` on my Ledger device while CLI shows the args to be `{"args": {}}` I would be confused and won't sign it until I understand it. If I learn that the args are whatever and no way to review them (I won't type the base64 from Ledger device to base64 decoder), I will put myself into a habit of blind signing and then it is a matter of time before https://blockaid.io/blog/how-to-prevent-the-next-15b-bybit-hack-a-strategic-approach-to-solving-blind-signing will happen.

src/transaction_signature_options/submit_dao_proposal/mod.rs
frol: @frolvanya proposal bond [is configurable](https://github.com/near-daos/sputnik-dao-contract/blob/main/sputnikdao2/src/proposals.rs#L551), and can be set to zero. We may pull the proposal bond from the contract settings, but this is more complex than just letting to enter any amount.

src/common.rs
frolvanya: Ah, makes sense then

src/transaction_signature_options/submit_dao_proposal/mod.rs
vsavchyn-dev: Understandable, fixed in [a2c8622](https://github.com/vsavchyn-dev/near-cli-rs/commit/a2c8622cb6914271a8c7fb0951576bc0a5edc7d6)

src/transaction_signature_options/submit_dao_proposal/mod.rs
vsavchyn-dev: Fixed in [a2c8622](https://github.com/vsavchyn-dev/near-cli-rs/commit/a2c8622cb6914271a8c7fb0951576bc0a5edc7d6)

P.S. When reviewing documentation for SputnikDAO, it was [mentioned](https://github.com/near-daos/sputnik-dao-contract?tab=readme-ov-file#add-proposal) that `add_proposal` "requires a minimum 1 Ⓝ bond (attached deposit)." Also, I totally missed that bond is configurable in code. That is why there I enforced min of 1 NEAR deposit...

src/transaction_signature_options/submit_dao_proposal/mod.rs
vsavchyn-dev: Understandable, fixed in [a2c8622](https://github.com/vsavchyn-dev/near-cli-rs/commit/a2c8622cb6914271a8c7fb0951576bc0a5edc7d6)

src/common.rs
vsavchyn-dev: Fixed in [a2c8622](https://github.com/vsavchyn-dev/near-cli-rs/commit/a2c8622cb6914271a8c7fb0951576bc0a5edc7d6)

src/transaction_signature_options/submit_dao_proposal/dao_kind_arguments.rs
vsavchyn-dev: I was thinking about doing this, however, I tried to cover several things:
- `contract call-function as-transaction`, which gives us one `FunctionCall` action -> can be translated to `FunctionCall` proposal kind
- `token send-near`, which gives us `Transfer` action -> can be translated to `Transfer` proposal kind
- `token send-ft`, which gives us at most 2 `FunctionCall` actions -> hence should be inside single `FunctionCall` proposal kind (I am not sure how to detect user sending ft and how to extract the token name from `FunctionCall` action to convert it to `Transfer` dao proposal, so I went with current approach)
- `token send-nft`, which gives us `FunctionCall` action -> can be translated to `FunctionCall` proposal kind
- `transaction construct-transaction`, with several `FunctionCall` actions -> can be translated to a single `FunctionCall` proposal kind

However, if converting each action into individual `add_proposal`, wouldn't it mean that user need to write description/gas/deposit for each `add_proposal`? 'Cause it would be strange apply same description/gas/deposit to all actions.

frol: @vsavchyn-dev Oh, I did not suggest to break each function call action into a separate add_proposal, I meant that we can translate TRANSFER + FUNCTION_CALL + FUNCTION_CALL into: (1) add_proposal [transfer kind] (2) add_proposal [2 function calls]

frol: But this might be not a realistic or useful case in reality, so let's drop this idea

frolvanya: > But this might be not a realistic or useful case in reality, so let's drop this idea

We can create a separate issue for this. But as you said, I don't think this would be a common case

Since now we have `add_proposal` functionality, I'd say that it's better to add an option to vote for a proposal instead of trying to perfect `add_proposal` functionality

frol (approved):
@race-of-sloths score 13

vsavchyn-dev:
@race-of-sloths

PR #496, opened 2025-06-24
chore: make --save-contract-source-code-into return no `.git` dir and no build artifacts (minimal working snapshot)

dj8yfo:
checked with absolute and relative paths
```bash
near contract verify --save-contract-source-code-into /home/jerrick/Documents/code/near/testing_little_simple_package deployed-at simple-package-verify-rs-ci.testnet network-config testnet now
# roundtrip with following is fine
near contract verify --use-contract-source-code-path /home/jerrick/Documents/code/near/testing_little_simple_package deployed-at simple-package-verify-rs-ci.testnet network-config testnet now

near contract verify --save-contract-source-code-into ./simple-pkg deployed-at simple-package-verify-rs-ci.testnet network-config testnet now
# roundtrip with following is fine
near contract verify --use-contract-source-code-path ./simple-pkg deployed-at simple-package-verify-rs-ci.testnet network-config testnet now
```


src/commands/contract/verify/mod.rs
```diff
@@ -235,6 +233,21 @@ fn get_contract_properties_from_docker_build(
     let target_dir = camino::Utf8PathBuf::from_path_buf(target_dir)
         .map_err(|err| color_eyre::eyre::eyre!("convert path buf {:?}", err))?;
 
+    if let Some(save_to_path) = save_contract_source_code_into {
+        // remove .git in target dir
+        let git_subfolder = target_dir.join(".git");
+        std::fs::remove_dir_all(&git_subfolder).wrap_err(format!(
+            "couldn't completely delete .git subfolder {:?}",
+            git_subfolder
+        ))?;
```
frol, line 241: As a developer, I wouldn't find it intuitive that with adding some flag the build folder will suddenly loose `.git` folder.

You may argue that SourceScan does snapshots to IPFS without `.git`, but I would argue that reproducible builds metadata does not point to IPFS and it is a "bug" of SourceScan. We can either:

1. Always remove `.git` folder for reproducible builds - this change must be tested for backward compatibility - near-sdk-rs `#[near]` macro that adds `contract_source_metadata()` relies on some git metadata that cargo provides out of the box as far as I remember
2. Say that SourceScan's IPFS is there only for source code reference/backup

Cargo.toml
```diff
@@ -116,6 +116,7 @@ near-verify-rs = {version = "0.3.0", optional = true}
 camino = {version = "1.1.1", optional = true}
 git2 = {version = "0.19", optional = true}
 tempfile = {version = "3.10.1", optional = true}
+dircpy = { version = "0.3.19", optional = true }
```
frol, line 119: I don't like this crate pulling in crossbeam, rayon and other fairly heavy dependencies for no good reason.

Let's either:
1. use [`fs_extra`](https://crates.io/crates/fs_extra), which is much more popular and has 0 dependencies
2. use git2 to clone from the locally cloned repo

src/commands/contract/verify/mod.rs
dj8yfo: Ok, agreed. Full clean git snapshot will be better from any perspective. 
This variant was just simpler to do in 2 lines as one of possible clean snapshots variants

PR #503, opened 2025-07-07
fix: use re-rexported version of `indicatif` module, needed for `tracing-indicatif`

dj8yfo:
image

https://github.com/emersonford/tracing-indicatif/blob/main/src/lib.rs#L46

---

https://github.com/actions/runner-images/issues/12045


dj8yfo:
@race-of-sloths include

src/common.rs
```diff
@@ -239,6 +239,7 @@ pub async fn get_account_transfer_allowance(
     })
 }
 
+#[allow(clippy::result_large_err)]
```
dj8yfo, line 242: TODO(depends on merge of this pr): create an issue to handle  this clippy lint  properly

PR #504, opened 2025-07-07
chore: add `verify_contract`/`inspect_contract` feauture gates

dj8yfo:


src/common.rs
```diff
@@ -2731,6 +2731,7 @@ pub(crate) fn teach_me_call_response(response: &impl serde::Serialize) {
     }
 }
 
+#[cfg(feature = "inspect_contract")]
 pub(crate) fn teach_me_request_payload(
```
dj8yfo, line 2732: this is only used in contract/inspect/mod.rs so maybe it should be moved there

src/commands/contract/mod.rs
```diff
@@ -57,3 +78,73 @@ pub enum ContractActions {
...
+    AbiNotSupported,
+    #[error("The contract has unknown NEAR ABI format (https://github.com/near/abi), so there is no way to get details about the function argument and return values. See more details about the error:\n\n{0}")]
+    AbiUnknownFormat(Report),
+    #[error("'__contract_abi' function call failed due to RPC error, so there is no way to get details about the function argument and return values. See more details about the error:\n\n{0}")]
+    RpcError(
+        near_jsonrpc_client::errors::JsonRpcError<
+            near_jsonrpc_primitives::types::query::RpcQueryError,
+        >,
+    ),
+}
```
dj8yfo, line 138: these 2 were moved up 1 level in module tree, as they're used in 3 commands: inspect, download_abi and call_function

dj8yfo:
@race-of-sloths include

frol (approved):
@race-of-sloths score 3

PR #510, opened 2025-07-21
feat: Updated transaction status handler (Started, NotStarted)

FroVolod:
This should not have been needed, but due to a bug on nearcore side, we have to handle the "Started" status gracefully: https://github.com/near/nearcore/issues/13872


frol (approved):
@race-of-sloths score 2

PR #513, opened 2025-08-07
fix: Fixed the inspection command for the global contract on the account.

### LINKED ISSUE #512: The CLI crashes on attempt to inspect an account's contract
Simply run the following command, and you're gonna see the error:
`near contract inspect "1754245704718.monkey-fruit-test.testnet" network-config testnet now`

Image

P.S. this account `1754245704718.monkey-fruit-test.testnet` has sent a transaction with the action `UseGlobalContract(CodeHash)` shortly before that, and since for other accounts inspection works fine, I would assume that it's somehow connected to Global Contracts

### Details

Target: `aarch64-apple-darwin`
Version: `0.22.0`
OS: MacOS Sonoma 14.4

FroVolod:
Resolves #512

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.22.1](https://github.com/near/near-cli-rs/compare/v0.22.0...v0.22.1) - 2025-08-15
+
+### Other
+
+- update near-* dependencies to 0.31 release ([#514](https://github.com/near/near-cli-rs/pull/514))
+
 ## [0.22.0](https://github.com/near/near-cli-rs/compare/v0.21.0...v0.22.0) - 2025-07-21
 
 ### Added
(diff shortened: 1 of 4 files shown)


src/commands/contract/inspect/mod.rs
```diff
@@ -157,15 +157,19 @@ async fn display_inspect_contract(
...
     table.add_row(prettytable::row![
         Fy->"Storage used",
-        format!("{} ({} Wasm + {} data)",
-            bytesize::ByteSize(account_view.storage_usage),
-            bytesize::ByteSize(u64::try_from(contract_code_view.code.len())?),
-            bytesize::ByteSize(
-                account_view.storage_usage
-                    .checked_sub(u64::try_from(contract_code_view.code.len())?)
-                    .expect("Unexpected error")
+        if account_view.global_contract_account_id.is_none() & account_view.global_contract_hash.is_none() {
```
akorchyn: Maybe we can somehow omit those `view_code_response` and fetch data only if necessary.
I see that it is provided as an argument, but can you take a look if it is possible to move this calculation here if it is not mandatory there.

FroVolod: done (519c98e)

frol (approved):

Issue #62, opened 2021-09-10
Tests against command output


ChaoticTempest:
With a lot of changes hitting in the future, testing will prove vital to making sure we get all of them working properly.

Currently output for commands are being printed to stdout which is really hard to test against. Some potential solutions could include just returning it and allowing some tests to be ran against them. Each of the commands from what I've seen just return `Result<()>` AFAIK.

This could be a large undertaking though, due to the amount of types that are present. Hopefully after reducing the boilerplate and deduplicating several types like `Sender`, we can get to a point where we can do this more easily.

PR #521, opened 2025-10-28
feat: reconstruct `deploy-global-contract`

frolvanya:
Should be merged after:
https://github.com/near/near-jsonrpc-client-rs/pull/179
https://github.com/bos-cli-rs/near-socialdb-client-rs/pull/36

image

### DIFF
--- a/src/commands/contract/download_wasm/mod.rs
+++ b/src/commands/contract/download_wasm/mod.rs
@@ -260,8 +260,8 @@ impl From<DownloadGlobalContractByAccountIdContext> for ArgsForDownloadContract 
 #[interactive_clap(input_context = crate::GlobalContext)]
 #[interactive_clap(output_context = DownloadGlobalContractByCodeHashContext)]
 pub struct DownloadGlobalContractByCodeHash {
-    /// What is the contract hash of the global contract?
-    contrach_hash: crate::types::crypto_hash::CryptoHash,
+    /// What is the code hash of the global contract?
+    code_hash: crate::types::crypto_hash::CryptoHash,
     #[interactive_clap(named_arg)]
     save_to_file: DownloadContract,
 }
@@ -279,7 +279,7 @@ impl DownloadGlobalContractByCodeHashContext {
     ) -> color_eyre::eyre::Result<Self> {
         Ok(Self {
             global_context: previous_context,
-            code_hash: scope.contrach_hash,
+            code_hash: scope.code_hash,
         })
     }
 }


src/commands/contract/download_wasm/mod.rs
```diff
@@ -124,24 +130,34 @@ fn download_contract_code(
 
 #[tracing::instrument(name = "Download contract code ...", skip_all)]
 pub fn get_code(
+    is_global_contract: bool,
```
frolvanya: This looks very ugly, but I'm not sure how to decouple this properly

1) `get_code` is used by both transaction reconstruction and by `download_wasm`. Also, I'm not sure why there're two different methods to download wasm file (one in reconstruct-transaction module and the other in download-wasm module)
2) I'm not sure if asking if the contract is global before downloading is the right way 

Maybe it's better to try to query `view_code` and then `view_global_contract_code_by_account_id` if the first query failed?

frolvanya: Mitigated by `ContractType` for now: https://github.com/near/near-cli-rs/pull/521/commits/6ebf18cf22064526f41398e6281c8abceb32cf96

Let me know if there's a better option to do this

src/commands/transaction/reconstruct_transaction/mod.rs
```diff
@@ -265,8 +266,33 @@ fn action_transformation(
...
             panic!("Internal error: Delegate action should have been handled before calling action_transformation.");
         }
-        Action::DeployGlobalContract(_) => {
-            Err(color_eyre::eyre::eyre!("Reconstruction of Global Deploy transactions is not supported yet. This feature is being tracked at: https://github.com/near/nearcore/issues/13531"))
+        Action::DeployGlobalContract(action) => {
+            download_code(
+                crate::commands::contract::download_wasm::ContractType::Global,
+                &receiver_id,
+                network_config,
+                near_primitives::types::BlockReference::latest(),
```
frolvanya: This is a workaround. Contract hash is not accessible at the block height with deploy-global transaction

Example:
image

https://testnet.nearblocks.io/txns/7iScLbaNApB47WwMhANr8NnddZaRzmNXHRqd5D2HjAnd

Contract was deployed at block 220576516, but code wasn't accessible until 220576518. Is it expected? What will be the best fix? Is it reliable to add always 2 blocks or does it fluctuate?

frol: > Is it expected?

Yes, transaction is first converted into a receipt, and the receipt is executed in the following blocks.

> Is it reliable to add always 2 blocks or does it fluctuate?

Nope, we need to use the block where the receipt execution outcomes are recorded. Check the tx status output for the receipt_outcomes, and use the block hash of the first receipt outcome.

src/commands/contract/download_wasm/mod.rs
```diff
@@ -124,24 +153,35 @@ fn download_contract_code(
...
 #[tracing::instrument(name = "Download contract code ...", skip_all)]
 pub fn get_code(
+    contract_type: ContractType,
     account_id: &near_primitives::types::AccountId,
     network_config: &crate::config::NetworkConfig,
     block_reference: near_primitives::types::BlockReference,
 ) -> color_eyre::eyre::Result<Vec<u8>> {
+    let request = match contract_type {
+        ContractType::Global => {
+            near_primitives::views::QueryRequest::ViewGlobalContractCodeByAccountId {
```
frol: Can you also add the support for the global contract deployed by code hash?

frol (changes requested):
Please, add support for global contracts deployed by code hash and address the other comment

src/commands/transaction/reconstruct_transaction/mod.rs
frolvanya: > Check the tx status output for the receipt_outcomes, and use the block hash of the first receipt outcome.

The only receipt_outcome in the transaction has block height ...516. I've sent a message to protocol team to get some clarity

frolvanya: I'll stick around with the option of simple loop to go through each block trying to fetch code

More information from the protocol team:

"there is at most 1 global contract distribution receipt in flight per global contract. Once a distribution receipt is processed on a destination shard, it checks on which shards the global contract was not distributed yet and creates a new receipt for one of those shards. The process keeps being repeated until all shards have the global contract, so the distribution can span multiple blocks"

"the distribution receipts are executed outside of the normal action receipt flow, so they don't have outcomes"

"qureying seems like the best option. In 99% of cases the contract should be distributed after `num_shards` blocks"

src/commands/contract/download_wasm/mod.rs
frolvanya: Added in https://github.com/near/near-cli-rs/pull/521/commits/ed889ff1fe337cdfae686718eb8e09c79ddfc59a

src/commands/transaction/reconstruct_transaction/mod.rs
frolvanya: Fixed in https://github.com/near/near-cli-rs/pull/521/commits/ed889ff1fe337cdfae686718eb8e09c79ddfc59a

src/commands/contract/download_wasm/mod.rs
```diff
@@ -1,26 +1,76 @@
...
     /// Enter the name of the file to save the contract:
     save_to_file: DownloadContract,
 }
 
+impl Contract {
+    pub fn input_contract_kind(
+        _context: &crate::GlobalContext,
+    ) -> color_eyre::eyre::Result<Option<ContractKind>> {
+        #[derive(strum_macros::Display, PartialEq)]
+        enum Options {
```
frolvanya: This seems like a workaround to me. I wanted to preserve `kebab-case` format for `global-contract-by-account-id` and other options in cli mode, but provide more readable display format in interactive mode

Let me know if there's a better way to do it instead of copying the same structure all over the place

PolyProgrammist (changes requested):
Hi! What can be my problem here:
```
polyprogrammist@Macbooks-MacBook-Pro-2 near-cli-rs % ./target/debug/near
> What are you up to? (select one of the options with the up-down arrows on your
 keyboard and press Enter) transaction - Operate transactions
> Сhoose action for transaction: reconstruct-transaction  - Use any existing tra
nsaction from the chain to construct NEAR CLI command (helpful tool for re-submi
tting similar transactions)
> Enter the hash of the transaction you want to use as a template: 7iScLbaNApB47
WwMhANr8NnddZaRzmNXHRqd5D2HjAnd
> What is the name of the network? fast
 INFO Transaction 7iScLbaNApB47WwMhANr8NnddZaRzmNXHRqd5D2HjAnd:
 |    signer_id:    testglobal.testnet
 |    receiver_id:  testglobal.testnet
 |    actions:   -- deploy code <AgCwqHHwiT2MsmKDeoCvc8K5w6dqsbmpXJZ9yX68kU2i> to a global account <testglobal.testnet>

 INFO Trying to fetch contract code for <testglobal.testnet> at block height 220576516 on network <testnet>...
 INFO Trying to fetch contract code for <testglobal.testnet> at block height 220576517 on network <testnet>...
 INFO Trying to fetch contract code for <testglobal.testnet> at block height 220576518 on network <testnet>...
 INFO Trying to fetch contract code for <testglobal.testnet> at block height 220576519 on network <testnet>...
 INFO Trying to fetch contract code for <testglobal.testnet> at block height 220576520 on network <testnet>...
 INFO Trying to fetch contract code for <testglobal.testnet> at block height 220576521 on network <testnet>...
 INFO Trying to fetch contract code for <testglobal.testnet> at block height 220576522 on network <testnet>...
 INFO Trying to fetch contract code for <testglobal.testnet> at block height 220576523 on network <testnet>...
 INFO Trying to fetch contract code for <testglobal.testnet> at block height 220576524 on network <testnet>...
 INFO Trying to fetch contract code for <testglobal.testnet> at block height 220576525 on network <testnet>...
 INFO Trying to fetch contract code for <testglobal.testnet> at block height 220576526 on network <testnet>...
 INFO Trying to fetch contract code for <testglobal.testnet> at block height 220576527 on network <testnet>...
 INFO Trying to fetch contract code for <testglobal.testnet> at block height 220576528 on network <testnet>...
 INFO Trying to fetch contract code for <testglobal.testnet> at block height 220576529 on network <testnet>...
 INFO Trying to fetch contract code for <testglobal.testnet> at block height 220576530 on network <testnet>...
 INFO Trying to fetch contract code for <testglobal.testnet> at block height 220576531 on network <testnet>...
 INFO Trying to fetch contract code for <testglobal.testnet> at block height 220576532 on network <testnet>...
 INFO Trying to fetch contract code for <testglobal.testnet> at block height 220576533 on network <testnet>...
 INFO Trying to fetch contract code for <testglobal.testnet> at block height 220576534 on network <testnet>...

Here is your console command if you need to script it or re-run:
    ./target/debug/near transaction reconstruct-transaction 7iScLbaNApB47WwMhANr8NnddZaRzmNXHRqd5D2HjAnd network-config fast

Error:
   0: Couldn't fetch the code. Please verify that you are using the archival node in the `network_connection.*.rpc_url` field of the `config.toml` file. You can see the list of RPC providers at https://docs.near.org/api/rpc/providers.
      Error: Failed to fetch contract code for <testglobal.testnet> on network <testnet> after trying 19 block heights.

Location:
   src/commands/transaction/reconstruct_transaction/mod.rs:403

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
polyprogrammist@Macbooks-MacBook-Pro-2 near-cli-rs %
```

I use the following config:
```
[network_connection.fast]
network_name = "testnet"
rpc_url = "https://archival-rpc.testnet.fastnear.com/"
wallet_url = "https://archival-rpc.testnet.fastnear.com/"
explorer_transaction_url = "https://archival-rpc.testnet.fastnear.com/"
```

src/commands/contract/download_wasm/mod.rs
PolyProgrammist: Please check how is it done for `FunctionArgsTypeDiscriminants`. Can the same be applied here?

src/commands/transaction/reconstruct_transaction/mod.rs
```diff
@@ -265,8 +266,33 @@ fn action_transformation(
         Action::Delegate(_) => {
             panic!("Internal error: Delegate action should have been handled before calling action_transformation.");
         }
-        Action::DeployGlobalContract(_) => {
-            Err(color_eyre::eyre::eyre!("Reconstruction of Global Deploy transactions is not supported yet. This feature is being tracked at: https://github.com/near/nearcore/issues/13531"))
+        Action::DeployGlobalContract(action) => {
+            download_code(
+                crate::commands::contract::download_wasm::ContractKind::GlobalContractByAccountId,
```
PolyProgrammist:
```suggestion
                crate::commands::contract::download_wasm::ContractKind,
```

src/commands/transaction/reconstruct_transaction/mod.rs
```diff
@@ -358,7 +384,8 @@ fn get_access_key_permission(
 }
 
 fn download_code(
-    receiver_id: &near_primitives::types::AccountId,
+    contract_kind: crate::commands::contract::download_wasm::ContractKind,
+    target: &str,
```
PolyProgrammist: Let's make it typed, not str. It's better to have a contract_kind and target together in a single struct. For example:
```
pub enum ContractKind {
    Regular(AccountId),
    GlobalContractByAccountId(AccountId),
    GlobalContractByHash(CryptoHash),
}
```

src/commands/contract/download_wasm/mod.rs
```diff
@@ -1,26 +1,76 @@
...
 
+use clap::ValueEnum;
 use color_eyre::eyre::Context;
 use inquire::CustomType;
 
 use crate::common::JsonRpcClientExt;
 
+#[derive(Debug, strum::Display, PartialEq, ValueEnum, Clone, Copy)]
+#[strum(serialize_all = "kebab_case")]
+pub enum ContractKind {
```
PolyProgrammist, line 40: Why doesn't it use `interactive_clap`?

src/commands/contract/download_wasm/mod.rs
```diff
@@ -122,33 +221,95 @@ fn download_contract_code(
     Ok(())
 }
 
-#[tracing::instrument(name = "Download contract code ...", skip_all)]
+#[tracing::instrument(name = "Trying to download contract code ...", skip_all)]
 pub fn get_code(
-    account_id: &near_primitives::types::AccountId,
+    contract_kind: ContractKind,
+    target: &str,
```
PolyProgrammist: same: make typed, not str

src/commands/contract/download_wasm/mod.rs
```diff
@@ -58,6 +111,46 @@ pub struct DownloadContract {
...
+                        &context.global_context.config.credentials_home_dir,
+                        "What is the global contract account ID?",
+                    )?
+                else {
+                    return Ok(None);
+                };
+                account_id.to_string()
+            }
+            ContractKind::GlobalContractByHash => {
+                CustomType::<near_primitives::hash::CryptoHash>::new(
```
PolyProgrammist: Looks like you don't have to do that if you use interactive_clap

src/commands/transaction/reconstruct_transaction/mod.rs
```diff
@@ -265,8 +266,33 @@ fn action_transformation(
...
         }
-        Action::DeployGlobalContract(_) => {
-            Err(color_eyre::eyre::eyre!("Reconstruction of Global Deploy transactions is not supported yet. This feature is being tracked at: https://github.com/near/nearcore/issues/13531"))
+        Action::DeployGlobalContract(action) => {
+            download_code(
+                crate::commands::contract::download_wasm::ContractKind::GlobalContractByAccountId,
+                receiver_id.as_str(),
+                network_config,
+                block_reference,
+                "reconstruct-transaction-deploy-code.wasm",
```
PolyProgrammist: Let's ask the user for the output and make "reconstruct-transaction-deploy-code.wasm" default variant

src/commands/contract/download_wasm/mod.rs
```diff
@@ -50,6 +100,9 @@ impl Contract {
 #[interactive_clap(input_context = ContractContext)]
 #[interactive_clap(output_context = DownloadContractContext)]
 pub struct DownloadContract {
```
PolyProgrammist: So you are modifying `download_wasm`. Is it possible to support `download_abi` as well as other methods? Are they working right now for global contracts?
Anyway, maybe that's for a different PR.

src/commands/contract/download_wasm/mod.rs
```diff
@@ -90,26 +191,24 @@ impl From<DownloadContractContext> for crate::network_view_at_block::ArgsForView
...
-        context: &ContractContext,
+        _context: &ContractContext,
     ) -> color_eyre::eyre::Result<Option<crate::types::path_buf::PathBuf>> {
         Ok(Some(
             CustomType::new("Enter the name of the file to save the contract:")
-                .with_starting_input(&format!(
-                    "{}.wasm",
-                    context.account_id.as_str().replace('.', "_")
-                ))
+                .with_starting_input("config.wasm")
```
PolyProgrammist: Let's add info about filename in context

src/commands/contract/download_wasm/mod.rs
```diff
@@ -122,33 +221,95 @@ fn download_contract_code(
...
-            query_view_method_response.kind
-        {
-            result
-        } else {
-            return Err(color_eyre::Report::msg("Error call result".to_string()));
+        })?
+        .shard_layout
+        .num_shards();
+
+    for block_height in block_height..=block_height + number_of_shards * 2 {
```
PolyProgrammist, line 380: I would not do that. If it can't be downloaded, that means the contract is not deployed yet. And the actual deployment moment is when it's already deployed on all the shards. If the user wants to get the code, they can use another block height

src/commands/transaction/reconstruct_transaction/mod.rs
frolvanya: I was trying to do it in a separate commit which was deprecated later, because I couldn't make it work properly in both cli and interactive mode. But I can try one more time

src/commands/contract/download_wasm/mod.rs
akorchyn: Abi is actually not an own JSONRPC call, but a function call. It means that you can only use it on instantiated calls. 

This is a kinda flaw in design, I think. Maybe we could chat with the Nearcore team, and it might make sense to allow some read-only calls to the global contracts.

src/commands/contract/download_wasm/mod.rs
frolvanya: User provides tx hash for a reconstruction, not a block. Please see https://github.com/near/near-cli-rs/pull/521#discussion_r2470089532 for more details why I choose this approach

src/commands/transaction/reconstruct_transaction/mod.rs
frolvanya: @PolyProgrammist Now I remembered what was the issue. I cannot use non-unit enum variants with `EnumDiscriminant`
image

PolyProgrammist: Can it be something like that?
```rust

use crate::types::crypto_hash::CryptoHash;
use crate::types::account_id::AccountId;
use strum::{EnumDiscriminants, EnumMessage, EnumIter};


#[derive(Debug, Clone)]
pub struct ContractTypeContext {
    config: crate::config::Config,
}

impl ContractTypeContext {
    pub fn from_previous_context(
        previous_context: crate::GlobalContext,
        scope: &<ContractType as interactive_clap::ToInteractiveClapContextScope>::InteractiveClapContextScope,
        Ok(Self {
            config: previous_context.config,
        })
    }
}

#[derive(Debug, EnumDiscriminants, Clone, interactive_clap::InteractiveClap)]
#[interactive_clap(input_context = crate::GlobalContext)]
#[interactive_clap(output_context = ContractTypeContext)]
#[strum_discriminants(derive(EnumMessage, EnumIter))]
pub enum ContractType {
    #[strum_discriminants(strum(
        message = "regular   - Regular contract"
    ))]
    Regular(GetAccountId),
    #[strum_discriminants(strum(
        message = "global-contract-by-account-id - Global contract by account ID"
    ))]
    GlobalContractByAccountId(GetAccountId),
    #[strum_discriminants(strum(
        message = "global-contract-by-hash - Global contract by hash"
    ))]
    GlobalContractByHash(GetCryptoHash),
}


#[derive(Debug, Clone, interactive_clap::InteractiveClap)]
#[interactive_clap(input_context = ContractTypeContext)]
pub struct GetAccountId {
    #[interactive_clap(skip_default_input_arg)]
    /// Account ID
    account_id: crate::types::account_id::AccountId,
}

impl GetAccountId {
    pub fn input_account_id(
        context: &ContractTypeContext,
    ) -> color_eyre::eyre::Result<Option<crate::types::account_id::AccountId>> {
        crate::common::input_signer_account_id_from_used_account_list(
            &context.config.credentials_home_dir,
            "Which account do you want to add an access key to?",
        )
    }
}

#[derive(Debug, Clone, interactive_clap::InteractiveClap)]
#[interactive_clap(input_context = ContractTypeContext)]
pub struct GetCryptoHash {
    #[interactive_clap(skip_default_input_arg)]
    /// Code hash
    code_hash: crate::types::crypto_hash::CryptoHash,
}

impl GetCryptoHash {
    pub fn input_code_hash(
        _context: &ContractTypeContext,
    ) -> color_eyre::eyre::Result<Option<crate::types::crypto_hash::CryptoHash>> {
        use inquire::CustomType;

        let code_hash = CustomType::<CryptoHash>::new("What is the contract code hash?").prompt()?;
        Ok(Some(code_hash))
    }
}

```

src/commands/contract/download_wasm/mod.rs
PolyProgrammist: Got it, thanks for clarifying

PolyProgrammist: Maybe the user can optionally specify the block number when it's distributed already. In case if the default behaviour with for cycle doesn't work. But I think it's ok wthout that

src/commands/transaction/reconstruct_transaction/mod.rs
frolvanya: Then `ContractType` becomes a subcommand in `Contract`? If this is a case, this would introduce code duplication to handle each contract type properly, because `save_to_file` is a `named_arg` which is a simplified version of the subcommand, but you can't have multiple subcommands in one structure.

Sure having a string is not the best, but we parse it properly here and it's not the first place in near-cli-rs that uses string instead of proper typing (I'm not advocating for a bad code quality, I just don't have enough experience with `interactive_clap` to make it beautiful)
image

src/commands/contract/download_wasm/mod.rs
frolvanya: In this case probably it's better to do it in a separate PR

frolvanya:
> Could you please add tests for your feature if it's not too hard? Otherwise if it's too big - then there is an issue for that #62, can be done in a separate pull request later

Not sure what exactly needs to be tested. Should I test that transaction reconstruction and download action download proper wasm at the certain block? 

Maybe I'm missing something, but I see tests only for types, not for action logic itself. Are there any examples of this?

src/commands/contract/download_wasm/mod.rs
frolvanya: User can use `download-wasm` with a specific block in this case. But protocol team said that it's very unlikely that contract distribution will take longer than the number of shards, so I took `num_shards * 2` as a safe bet

src/commands/contract/download_wasm/mod.rs
frolvanya: Like this: f450da17b9a28af0082f0af82102be0c2f54ea0e?

src/commands/contract/download_wasm/mod.rs
frolvanya: Fixed in 1b29ad660f6f309d1bce9d519f9339f069d078bd +  c0c464858747b1e0f94818fcb44c5e9ee3231265

src/commands/contract/download_wasm/mod.rs
frolvanya: Fixed in 1b29ad660f6f309d1bce9d519f9339f069d078bd

src/commands/transaction/reconstruct_transaction/mod.rs
frolvanya: Added in 211900c878fa8ee11d9d4ba90380450f5a49eb34

src/commands/transaction/reconstruct_transaction/mod.rs
frolvanya: I don't quite understand this comment. Could you please clarify why we need to remove the variant and pass enum?

src/commands/transaction/reconstruct_transaction/mod.rs
frolvanya: @PolyProgrammist maybe you know any workarounds to solve this without duplicating download context for both GetAccountId and GetCryptoHash?

image

Context: 
```rust
#[derive(Debug, Clone, interactive_clap::InteractiveClap)]
#[interactive_clap(input_context = ContractTypeContext)]
pub struct GetAccountId {
    #[interactive_clap(skip_default_input_arg)]
    /// Account ID
    account_id: crate::types::account_id::AccountId,
    #[interactive_clap(named_arg)]
    /// Enter the name of the file to save the contract:
    save_to_file: DownloadContract,
}

#[derive(Debug, Clone, interactive_clap::InteractiveClap)]
#[interactive_clap(input_context = ContractTypeContext)]
pub struct GetCryptoHash {
    #[interactive_clap(skip_default_input_arg)]
    /// Code hash
    code_hash: crate::types::crypto_hash::CryptoHash,
    #[interactive_clap(named_arg)]
    /// Enter the name of the file to save the contract:
    save_to_file: DownloadContract,
}

#[derive(Debug, Clone, interactive_clap::InteractiveClap)]
#[interactive_clap(input_context = ???)] // We need to accept both GetAccountId and GetCryptoHash
#[interactive_clap(output_context = DownloadContractContext)]
pub struct DownloadContract {
    /// Account ID (regular/global-contract-by-account-id) or code hash (global-by-hash)
    #[interactive_clap(skip_default_input_arg)]
    /// Enter the name of the file to save the contract:
    file_path: crate::types::path_buf::PathBuf,
    #[interactive_clap(named_arg)]
    /// Select network
    network_config: crate::network_view_at_block::NetworkViewAtBlockArgs,
}


#[derive(Debug, Clone, interactive_clap::InteractiveClap)]
#[interactive_clap(input_context = crate::GlobalContext)]
#[interactive_clap(output_context = ContractContext)]
pub struct Contract {
    #[interactive_clap(subcommand)]
    /// Which type of contract do you want to download?
    contract_type: ContractType,
    
    // save_to_file: DownloadContract was moved to `ContractType` because we can't have both `named_arg` and `subcommand` in the same structure
}
```

src/commands/transaction/reconstruct_transaction/mod.rs
PolyProgrammist: There is no code that handles GlobalContractByHash as soon as I see

src/commands/transaction/reconstruct_transaction/mod.rs
frolvanya: Nope

src/commands/contract/download_wasm/mod.rs
PolyProgrammist: I am sorry, I guess I misunderstood what's happening here. Could you please revert that?

src/commands/transaction/reconstruct_transaction/mod.rs
PolyProgrammist: Please take a look at the example with `ExportAccountActions`. This enum has three options: `ExportAccountFromWebWallet`, `ExportAccountFromSeedPhrase` and `ExportAccountFromPrivateKey`. They all end up having
```
     #[interactive_clap(named_arg)]
    network_config: crate::network::Network
```

Is it possible here?

src/commands/contract/download_wasm/mod.rs
frolvanya: Reverted in fe464a2d6d9600ea8c66fa914239bb8c8951f4d2

src/commands/transaction/reconstruct_transaction/mod.rs
frolvanya: As far as I understand it's possible because when each structure (in our case `GetAccountId` and `GetCryptoHash`) has a context as a terminal step, which is not a case for us. In our case we have `DownloadContext` is a final output

I'm not sure if refactor with duplication of logic inside `DownloadContext` for both structures is worth it

cc @frol

src/commands/transaction/reconstruct_transaction/mod.rs
frolvanya: Makes sense. Can we take code hash from `action.code` like this?
2e0ebc0613951d3a846aa9547b1ed258d252c38f

src/commands/transaction/reconstruct_transaction/mod.rs
PolyProgrammist: I believe it can be refactored as a terminal step without any duplication. 

Though, the code is already quite solid. It works great and well-implemented. So we have to merge it so that the users can already enjoy the feature

PolyProgrammist:
@race-of-sloths score 11

PolyProgrammist (approved):

src/commands/contract/download_wasm/mod.rs
```diff
@@ -2,25 +2,102 @@ use std::io::Write;
...
+
 #[derive(Debug, Clone, interactive_clap::InteractiveClap)]
 #[interactive_clap(input_context = crate::GlobalContext)]
 #[interactive_clap(output_context = ContractContext)]
 pub struct Contract {
     #[interactive_clap(skip_default_input_arg)]
-    /// What is the contract account ID?
-    account_id: crate::types::account_id::AccountId,
+    /// Which type of contract do you want to download?
+    contract_kind: ContractKind,
```
frol: There is some serious overengineering here. Can you jsut turn `ContractKind` into a subcommand like:

https://github.com/near/near-cli-rs/blob/c9190cb677bb2c8c060aed698264d330d60c8edc/src/commands/mod.rs#L35-L39

https://github.com/near/near-cli-rs/blob/c9190cb677bb2c8c060aed698264d330d60c8edc/src/commands/contract/mod.rs#L19-L31

```rust
#[derive(..., interactive_clap::InteractiveClap)]
struct Contract {
    #[interactive_clap(subcommand)]
    contract_kind: ContractKind,
}

#[derive(..., interactive_clap::InteractiveClap)]
enum ContractKind {
    Regular(DownloadRegularContract),
    GlobalContractByAccountId(DownloadGlobalContractByAccountId),
    GlobalContractByContractHash(DownloadGlobalContractByContractHash),
}

#[derive(..., interactive_clap::InteractiveClap)]
struct DownloadRegularContract {
    account_id: ...
    #[interactive_clap(named_arg)]
    save_to_file: DownloadContract,
}

#[derive(..., interactive_clap::InteractiveClap)]
struct DownloadGlobalContractByAccountId {
    account_id: ...
    #[interactive_clap(named_arg)]
    save_to_file: DownloadContract,
}

#[derive(..., interactive_clap::InteractiveClap)]
struct DownloadGlobalContractByContractHash {
    contract_hash: ...
    #[interactive_clap(named_arg)]
    save_to_file: DownloadContract,
}
```

This way at the `DownloadRegularContract` and other two structs you can implement a context that will get the ContractKind discriminant in the scope, and pass it down in the context.

frolvanya: Again, we can't use DownloadContract as you provided in example, since context can't have multiple (3 in this case) different input contexts. This can be solved by creating a separate context to download contract as I mentioned from the start, but it involves code duplication

https://github.com/near/near-cli-rs/pull/521/commits/0d0d942b6985cafd2dfa58818d04ef822d6fdee5

frolvanya: Refactored in b6e49c6a4bc636aa27ec34af2fa0d0c5b6a2ff91

src/commands/contract/download_wasm/mod.rs
frol: @frolvanya The problem with this approach is that if there was already a global contract registerred before, this code will download the outdated contract. One solution I see is to also pass `code_hash: Option<CryptoHash>` to verify that the downloaded contract actually matches the one in the transaction and keep iterating over the block height until we download the code with the matching hash

src/commands/transaction/reconstruct_transaction/mod.rs
```diff
@@ -211,18 +212,22 @@ fn action_transformation(
             )))
         }
         Action::DeployContract(deploy_contract_action) => {
+            let file_path = CustomType::<String>::new("Enter the name of the file to save the contract:")
```
frol: Let's use `PathBuf` instead of `String`

src/commands/transaction/reconstruct_transaction/mod.rs
```diff
@@ -265,8 +270,46 @@ fn action_transformation(
         Action::Delegate(_) => {
             panic!("Internal error: Delegate action should have been handled before calling action_transformation.");
         }
-        Action::DeployGlobalContract(_) => {
-            Err(color_eyre::eyre::eyre!("Reconstruction of Global Deploy transactions is not supported yet. This feature is being tracked at: https://github.com/near/nearcore/issues/13531"))
+        Action::DeployGlobalContract(action) => {
+            let file_path = CustomType::<String>::new("Enter the name of the file to save the contract:")
```
frol: Also `PathBuf`

frol (commented):
@frolvanya Now it looks good. Thank you for iterating on it!

src/commands/transaction/reconstruct_transaction/mod.rs
frolvanya: Fixed in cce2f8c8c8729d8fb3090a7cc9a66e3eb35414b2

src/commands/transaction/reconstruct_transaction/mod.rs
frolvanya: Fixed in cce2f8c8c8729d8fb3090a7cc9a66e3eb35414b2

src/commands/contract/download_wasm/mod.rs
frolvanya: Fixed in https://github.com/near/near-cli-rs/pull/521/commits/94acf30a38ae67ad2e831452d9373fcf5c871e19

src/common.rs
```diff
@@ -847,7 +847,8 @@ pub fn print_unsigned_transaction(
                 info_str.push_str(&print_unsigned_transaction(&prepopulated_transaction));
             }
             near_primitives::transaction::Action::DeployGlobalContract(deploy) => {
-                let code_hash = CryptoHash::hash_bytes(&deploy.code);
+                let code_hash = near_primitives::hash::CryptoHash::try_from(deploy.code.as_ref())
+                    .expect("The code hash must be valid since it is derived from the code bytes.");
```
frolvanya, line 851: Btw, we don't need to take a hash from hash, we just need to convert already hashed bytes into `CryptoHash`. I didn't check if this issue persists in other places, but probably it's not the only occurrence of this issue

frol: This is another example of unfortunate naming that got through the time 😢 

The original change:

https://github.com/near/nearcore/commit/a295e67097427a8450f02185380377d9f4ae90cf#diff-1e4fc99d32e48420a9bd37050fa1412758cba37825851edea40cbdfcab406944L550

The "solidified legacy":

https://github.com/near/nearcore/pull/12737#discussion_r1916374461

frol (approved):
@race-of-sloths score 8

PR #522, opened 2025-11-05
feat: introduce tests

PolyProgrammist:
It's not changing to result. But maybe just parsing output is good for e2e

Tests for 2 endpoints

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -19,7 +19,7 @@ pub struct GlobalContext {
     pub verbosity: Verbosity,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Default)]
 pub enum Verbosity {
     #[default]
     Interactive,


tests/common/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+    // Write a V3 config to avoid migration issues
+    let config_content = format!(
+        r#"version = "3"
+credentials_home_dir = "{}"
+
+[network_connection.sandbox]
+network_name = "sandbox"
+rpc_url = "{}"
+wallet_url = "{}"
+explorer_transaction_url = "{}transactions/"
```
frol, line 37: I’d just hardcode the wallet and explorer URLs (instead of abusing the RPC URL), so in the captured output the explorer links look more reasonable

frol (approved):
@race-of-sloths score 8

PR #531, opened 2025-12-17
fix: Fixed UI for interactive mode

FroVolod:


frol (approved):
@race-of-sloths score 5

PR #542, opened 2026-01-01
fix: Fixed testnet-lava RPC URL

frol:


frol:
@race-of-sloths score include

race-of-sloths:
### ❌ Self-scoring is not permitted. 
Ask your peers to review and score your contributions!

PR #560, opened 2026-02-27
feat: state-init command

pityjllk:
Add StateInit support for deterministic account creation (NEP-616)

  This PR adds support for creating deterministic accounts via DeterministicAccountStateInit as specified in https://github.com/near/NEPs/blob/master/neps/nep-0616.md.

  New command: contract state-init

  Creates a deterministic account by providing a state init. The state init can be specified in three ways:

  - use-global-hash <hash> — reference a globally deployed contract by its code hash
  - use-global-account-id <account_id> — reference a globally deployed contract by the account that deployed it
  - from-borsh-base64 <base64> — provide a pre-serialized DeterministicAccountStateInit directly

  For the first two modes, the initialization data (a key-value map with base64-encoded keys and values) is provided separately:

  - data-from-json '<json>' — inline JSON, e.g. data-from-json '{"AAEC": "AwQF"}'
  - data-from-file <path> — read the same JSON format from a file

  The receiver account ID is automatically derived from the state init contents (deterministic keccak256-based address with 0s prefix).

  Changes to transaction construct-transaction

  The receiver is now selected via a ReceiverMode enum:

  - account-id — specify the receiver account ID directly (existing behavior)
  - state-init — derive the receiver from a deterministic state init, using the same three input modes described above

  This was achieved by splitting the old ConstructTransactionContext into two stages: ConstructTransactionSenderContext (sender only) → ConstructTransactionContext (sender + receiver +
  actions). The receiver selection step sits between the two.

 Example usage

  Create a deterministic account and deploy via state-init:

```
  near transaction construct-transaction signer.near \
    state-init use-global-hash FaJXVgS82fXhrvvC8yXV4ibHujW63KvL7dVvJiZ9naga \
    data-from-json '{"":"EAAAAHBpdHlqbGxrLnRlc3RuZXQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAANiLmaE3mtdPRW1WVXs6pXbGdg5/4OmhjI6pswq6Nr/j"}' \
    deposit '0 NEAR' \
    add-action function-call gd_deploy \
      file-args /path/to/contract.wasm.borsh \
      prepaid-gas '300.0 Tgas' attached-deposit '14.5 NEAR' \
    skip network-config testnet sign-with-keychain send
```
  Call a function on a deterministic account (receiver specified directly):
```
  near transaction construct-transaction matdeployer.testnet \
    account-id 0s49b604786a4a44077ef5a450ddf59e90ae3d95d0 \
    add-action function-call gd_approve \
      json-args '{"old_hash":"d88b99a...","new_hash":"d88b99a..."}' \
      prepaid-gas '100.0 Tgas' attached-deposit '1 yoctoNEAR' \
    skip network-config testnet sign-with-keychain send
```
  The account-id keyword is optional — the old positional syntax still works:
```
  near transaction construct-transaction matdeployer.testnet \
    0s49b604786a4a44077ef5a450ddf59e90ae3d95d0 \
    add-action function-call gd_approve \
      json-args '{"old_hash":"d88b99a...","new_hash":"d88b99a..."}' \
      prepaid-gas '100.0 Tgas' attached-deposit '1 yoctoNEAR' \
    skip network-config testnet sign-with-keychain send
```
  Backward compatibility

  The old CLI syntax (construct-transaction <sender> <receiver> add-action ...) continues to work — it's treated as the
  account-id receiver variant. Snapshot tests verify that both old and new syntaxes produce identical transactions (same hash,
  same serialized bytes).

### DIFF
--- a/src/commands/contract/state_init/mod.rs
+++ b/src/commands/contract/state_init/mod.rs
@@ -12,22 +12,18 @@ pub struct StateInit {
 #[interactive_clap(context = crate::GlobalContext)]
 #[strum_discriminants(derive(EnumMessage, EnumIter))]
 #[non_exhaustive]
-/// How do you want to identify the global contract code?
 pub enum StateInitModeCommand {
     #[strum_discriminants(strum(
-        message = "use-global-hash       - Use a global contract code hash (immutable)"
+        message = "use-global-hash       - Use a global contract code hash"
     ))]
-    /// Use a global contract code hash (immutable)
     UseGlobalHash(StateInitWithContractHashRef),
     #[strum_discriminants(strum(
-        message = "use-global-account-id - Use a global contract account ID (mutable)"
+        message = "use-global-account-id - Use a global contract account ID"
     ))]
-    /// Use a global contract account ID (mutable)
     UseGlobalAccountId(StateInitWithContractRefByAccount),
     #[strum_discriminants(strum(
         message = "from-borsh-base64     - Provide the entire state init as a borsh+base64 blob"
     ))]
-    /// Provide the entire DeterministicAccountStateInit as a borsh-serialized + base64-encoded blob
     FromBorshBase64(StateInitFromBorshBase64),
 }
 
@@ -35,7 +31,6 @@ pub enum StateInitModeCommand {
 #[interactive_clap(input_context = crate::GlobalContext)]
 #[interactive_clap(output_context = StateInitWithContractHashRefContext)]
 pub struct StateInitWithContractHashRef {
-    /// What is the base58-encoded hash of the global contract?
     pub hash: crate::types::crypto_hash::CryptoHash,
     #[interactive_clap(subcommand)]
     data: Data,
@@ -45,7 +40,6 @@ pub struct StateInitWithContractHashRef {
 #[interactive_clap(input_context = crate::GlobalContext)]
 #[interactive_clap(output_context = StateInitWithContractRefByAccountContext)]
 pub struct StateInitWithContractRefByAccount {
-    /// What is the account ID of the global contract?
     pub account_id: crate::types::account_id::AccountId,
     #[interactive_clap(subcommand)]
     data: Data,
@@ -105,10 +99,8 @@ impl StateInitWithContractRefByAccountContext {
 #[interactive_clap(input_context = crate::GlobalContext)]
 #[interactive_clap(output_context = StateInitFromBorshBase64Context)]
 pub struct StateInitFromBorshBase64 {
-    /// Enter the borsh-serialized + base64-encoded DeterministicAccountStateInit blob:
     pub state_init_base64: String,
     #[interactive_clap(named_arg)]
-    /// Specify deposit
     deposit: Deposit,
 }
 
@@ -141,17 +133,14 @@ impl From<StateInitFromBorshBase64Context> for StateInitDataContext {
 #[interactive_clap(context = StateInitModeContext)]
 #[strum_discriminants(derive(EnumMessage, EnumIter))]
 #[non_exhaustive]
-/// How do you want to provide the initial state data?
 pub enum Data {
     #[strum_discriminants(strum(
         message = "data-from-file - Read base64-encoded key-value JSON data from a file"
     ))]
-    /// Read base64-encoded key-value JSON data from a file
     DataFromFile(DataFromFile),
     #[strum_discriminants(strum(
-        message = "data-from-json - Provide base64-encoded key-value JSON data inline"
+        message = "data-from-json - Provide base64-encoded key-value JSON data inline (e.g. '{\"AAEC\": \"AwQF\"})')"
     ))]
-    /// Provide base64-encoded key-value JSON data inline
     DataFromJson(DataFromJson),
 }
 
@@ -159,10 +148,8 @@ pub enum Data {
 #[interactive_clap(input_context = StateInitModeContext)]
 #[interactive_clap(output_context = DataFromFileContext)]
 pub struct DataFromFile {
-    /// What is the file path of the JSON state data?
     pub file_path: crate::types::path_buf::PathBuf,
     #[interactive_clap(named_arg)]
-    /// Specify deposit
     deposit: Deposit,
 }
 
@@ -170,10 +157,8 @@ pub struct DataFromFile {
 #[interactive_clap(input_context = StateInitModeContext)]
 #[interactive_clap(output_context = DataFromJsonContext)]
 pub struct DataFromJson {
-    /// Enter base64-encoded key-value JSON data (e.g. '{"AAEC": "AwQF"}' or '{}' for empty state):
     pub data: String,
     #[interactive_clap(named_arg)]
-    /// Specify deposit
     deposit: Deposit,
 }
 
@@ -263,13 +248,10 @@ impl DataFromJsonContext {
 #[interactive_clap(input_context = StateInitDataContext)]
 #[interactive_clap(output_context = DepositContext)]
 pub struct Deposit {
-    /// How much do you want to deposit with the state init (e.g. '1 NEAR' or '0 NEAR')?
     pub deposit: crate::types::near_token::NearToken,
     #[interactive_clap(skip_default_input_arg)]
-    /// What is the signer account ID?
     pub signer_account_id: crate::types::account_id::AccountId,
     #[interactive_clap(named_arg)]
-    /// Select network
     network_config: crate::network_for_transaction::NetworkForTransactionArgs,
 }
 
--- a/src/commands/contract/state_init/mod.rs
+++ b/src/commands/contract/state_init/mod.rs
@@ -23,9 +23,13 @@ pub enum StateInitModeCommand {
     ))]
     UseGlobalAccountId(StateInitWithContractRefByAccount),
     #[strum_discriminants(strum(
-        message = "from-borsh-base64     - Provide borsh serialized base64 encoded state init"
+        message = "from-borsh-base64      - Provide borsh serialized base64 encoded state init"
     ))]
     FromBorshBase64(StateInitFromBorshBase64),
+    #[strum_discriminants(strum(
+        message = "from-borsh-base64-file - Read borsh serialized base64 encoded state init from a file"
+    ))]
+    FromBorshBase64File(StateInitFromBorshBase64File),
 }
 
 #[derive(Debug, Clone, interactive_clap::InteractiveClap)]
@@ -141,6 +145,49 @@ impl StateInitFromBorshBase64Context {
 
 impl From<StateInitFromBorshBase64Context> for StateInitDataContext {
     fn from(item: StateInitFromBorshBase64Context) -> Self {
+        item.0
+    }
+}
+
+#[derive(Debug, Clone, interactive_clap::InteractiveClap)]
+#[interactive_clap(input_context = crate::GlobalContext)]
+#[interactive_clap(output_context = StateInitFromBorshBase64FileContext)]
+pub struct StateInitFromBorshBase64File {
+    /// Enter the path to a file containing borsh-base64 encoded StateInit:
+    pub file_path: crate::types::path_buf::PathBuf,
+    #[interactive_clap(named_arg)]
+    deposit: Deposit,
+}
+
+#[derive(Debug, Clone)]
+pub struct StateInitFromBorshBase64FileContext(StateInitDataContext);
+
+impl StateInitFromBorshBase64FileContext {
+    pub fn from_previous_context(
+        previous_context: crate::GlobalContext,
+        scope: &<StateInitFromBorshBase64File as interactive_clap::ToInteractiveClapContextScope>::InteractiveClapContextScope,
+    ) -> color_eyre::eyre::Result<Self> {
+        let base64_str = std::fs::read_to_string(&scope.file_path).wrap_err_with(|| {
+            format!(
+                "Failed to open or read the file: {}",
+                scope.file_path.0.display()
+            )
+        })?;
+        let data = near_primitives::serialize::from_base64(base64_str.trim())
+            .wrap_err("Failed to decode base64 from file content")?;
+        let state_init = crate::common::parse_borsh_base64_state_init(&data)?;
+        let receiver_account_id =
+            near_primitives::utils::derive_near_deterministic_account_id(&state_init);
+        Ok(Self(StateInitDataContext {
+            global_context: previous_context,
+            state_init,
+            receiver_account_id,
+        }))
+    }
+}
+
+impl From<StateInitFromBorshBase64FileContext> for StateInitDataContext {
+    fn from(item: StateInitFromBorshBase64FileContext) -> Self {
         item.0
     }
 }
--- a/src/commands/contract/state_init/mod.rs
+++ b/src/commands/contract/state_init/mod.rs
@@ -22,7 +22,7 @@ pub enum StateInitModeCommand {
     ))]
     UseGlobalAccountId(StateInitWithContractRefByAccount),
     #[strum_discriminants(strum(
-        message = "from-borsh-base64     - Provide the entire state init as a borsh+base64 blob"
+        message = "from-borsh-base64     - Provide borsh serialized base64 encoded state init"
     ))]
     FromBorshBase64(StateInitFromBorshBase64),
 }
@@ -350,6 +350,7 @@ impl From<DepositContext> for crate::commands::ActionContext {
             on_after_sending_transaction_callback: std::sync::Arc::new(
                 |_outcome_view, _network_config| Ok(()),
             ),
+            sign_as_delegate_action: false,
         }
     }
 }
--- a/src/commands/contract/state_init/mod.rs
+++ b/src/commands/contract/state_init/mod.rs
@@ -1,0 +1,332 @@
+use color_eyre::eyre::Context;
+use strum::{EnumDiscriminants, EnumIter, EnumMessage};
+
+#[derive(Debug, Clone, interactive_clap::InteractiveClap)]
+#[interactive_clap(context = crate::GlobalContext)]
+pub struct StateInit {
+    #[interactive_clap(subcommand)]
+    state_init: StateInitModeCommand,
+}
+
+#[derive(Debug, EnumDiscriminants, Clone, interactive_clap::InteractiveClap)]
+#[interactive_clap(context = crate::GlobalContext)]
+#[strum_discriminants(derive(EnumMessage, EnumIter))]
+#[non_exhaustive]
+/// How do you want to identify the global contract code?
+pub enum StateInitModeCommand {
+    #[strum_discriminants(strum(
+        message = "use-global-hash       - Use a global contract code hash (immutable)"
+    ))]
+    /// Use a global contract code hash (immutable)
+    UseGlobalHash(StateInitWithContractHashRef),
+    #[strum_discriminants(strum(
+        message = "use-global-account-id - Use a global contract account ID (mutable)"
+    ))]
+    /// Use a global contract account ID (mutable)
+    UseGlobalAccountId(StateInitWithContractRefByAccount),
+}
+
+#[derive(Debug, Clone, interactive_clap::InteractiveClap)]
+#[interactive_clap(input_context = crate::GlobalContext)]
+#[interactive_clap(output_context = StateInitWithContractHashRefContext)]
+pub struct StateInitWithContractHashRef {
+    /// What is the hash of the global contract?
+    pub hash: crate::types::crypto_hash::CryptoHash,
+    #[interactive_clap(subcommand)]
+    data: Data,
+}
+
+#[derive(Debug, Clone, interactive_clap::InteractiveClap)]
+#[interactive_clap(input_context = crate::GlobalContext)]
+#[interactive_clap(output_context = StateInitWithContractRefByAccountContext)]
+pub struct StateInitWithContractRefByAccount {
+    /// What is the account ID of the global contract?
+    pub account_id: crate::types::account_id::AccountId,
+    #[interactive_clap(subcommand)]
+    data: Data,
+}
+
+#[derive(Debug, Clone)]
+pub struct StateInitModeContext {
+    pub global_context: crate::GlobalContext,
+    pub code: near_primitives::action::GlobalContractIdentifier,
+}
+
+#[derive(Debug, Clone)]
+pub struct StateInitWithContractHashRefContext(StateInitModeContext);
+
+impl StateInitWithContractHashRefContext {
+    pub fn from_previous_context(
+        previous_context: crate::GlobalContext,
+        scope: &<StateInitWithContractHashRef as interactive_clap::ToInteractiveClapContextScope>::InteractiveClapContextScope,
+    ) -> color_eyre::eyre::Result<Self> {
+        Ok(Self(StateInitModeContext {
+            global_context: previous_context,
+            code: near_primitives::action::GlobalContractIdentifier::CodeHash(scope.hash.into()),
+        }))
+    }
+}
+
+impl From<StateInitWithContractHashRefContext> for StateInitModeContext {
+    fn from(item: StateInitWithContractHashRefContext) -> Self {
+        item.0
+    }
+}
+
+impl From<StateInitWithContractRefByAccountContext> for StateInitModeContext {
+    fn from(item: StateInitWithContractRefByAccountContext) -> Self {
+        item.0
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct StateInitWithContractRefByAccountContext(StateInitModeContext);
+
+impl StateInitWithContractRefByAccountContext {
+    pub fn from_previous_context(
+        previous_context: crate::GlobalContext,
+        scope: &<StateInitWithContractRefByAccount as interactive_clap::ToInteractiveClapContextScope>::InteractiveClapContextScope,
+    ) -> color_eyre::eyre::Result<Self> {
+        Ok(Self(StateInitModeContext {
+            global_context: previous_context,
+            code: near_primitives::action::GlobalContractIdentifier::AccountId(
+                scope.account_id.clone().into(),
+            ),
+        }))
+    }
+}
+
+#[derive(Debug, EnumDiscriminants, Clone, interactive_clap::InteractiveClap)]
+#[interactive_clap(context = StateInitModeContext)]
+#[strum_discriminants(derive(EnumMessage, EnumIter))]
+#[non_exhaustive]
+/// How do you want to provide the initial state data?
+pub enum Data {
+    #[strum_discriminants(strum(
+        message = "data-from-file - Read hex-encoded key-value JSON data from a file"
+    ))]
+    /// Read hex-encoded key-value JSON data from a file
+    DataFromFile(DataFromFile),
+    #[strum_discriminants(strum(
+        message = "data-from-json - Provide hex-encoded key-value JSON data inline"
+    ))]
+    /// Provide hex-encoded key-value JSON data inline
+    DataFromJson(DataFromJson),
+}
+
+#[derive(Debug, Clone, interactive_clap::InteractiveClap)]
+#[interactive_clap(input_context = StateInitModeContext)]
+#[interactive_clap(output_context = DataFromFileContext)]
+pub struct DataFromFile {
+    /// What is the file path of the JSON state data?
+    pub file_path: crate::types::path_buf::PathBuf,
+    #[interactive_clap(named_arg)]
+    /// Specify deposit
+    deposit: Deposit,
+}
+
+#[derive(Debug, Clone, interactive_clap::InteractiveClap)]
+#[interactive_clap(input_context = StateInitModeContext)]
+#[interactive_clap(output_context = DataFromJsonContext)]
+pub struct DataFromJson {
+    /// Enter hex-encoded key-value JSON data (e.g. '{"deadbeef": "cafebabe"}' or '{}' for empty state):
+    pub data: String,
+    #[interactive_clap(named_arg)]
+    /// Specify deposit
+    deposit: Deposit,
+}
+
+#[derive(Debug, Clone)]
+pub struct StateInitDataContext {
+    pub global_context: crate::GlobalContext,
+    pub state_init: near_primitives::deterministic_account_id::DeterministicAccountStateInit,
+    pub receiver_account_id: near_primitives::types::AccountId,
+}
+
+impl StateInitDataContext {
+    fn build(
+        code: near_primitives::action::GlobalContractIdentifier,
+        global_context: crate::GlobalContext,
+        data: std::collections::BTreeMap<Vec<u8>, Vec<u8>>,
+    ) -> color_eyre::eyre::Result<Self> {
+        let state_init =
+            near_primitives::deterministic_account_id::DeterministicAccountStateInit::V1(
+                near_primitives::deterministic_account_id::DeterministicAccountStateInitV1 {
+                    code,
+                    data,
+                },
+            );
+        let receiver_account_id =
+            near_primitives::utils::derive_near_deterministic_account_id(&state_init);
+        Ok(Self {
+            global_context,
+            state_init,
+            receiver_account_id,
+        })
+    }
+}
+
+impl From<DataFromFileContext> for StateInitDataContext {
+    fn from(item: DataFromFileContext) -> Self {
+        item.0
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct DataFromFileContext(StateInitDataContext);
+
+impl DataFromFileContext {
+    pub fn from_previous_context(
+        previous_context: StateInitModeContext,
+        scope: &<DataFromFile as interactive_clap::ToInteractiveClapContextScope>::InteractiveClapContextScope,
+    ) -> color_eyre::eyre::Result<Self> {
+        let json_str = std::fs::read_to_string(&scope.file_path).wrap_err_with(|| {
+            format!(
+                "Failed to open or read the file: {}",
+                scope.file_path.0.display()
+            )
+        })?;
+        let data = crate::common::parse_hex_kv_map(&json_str)?;
+        Ok(Self(StateInitDataContext::build(
+            previous_context.code,
+            previous_context.global_context,
+            data,
+        )?))
+    }
+}
+
+impl From<DataFromJsonContext> for StateInitDataContext {
+    fn from(item: DataFromJsonContext) -> Self {
+        item.0
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct DataFromJsonContext(StateInitDataContext);
+
+impl DataFromJsonContext {
+    pub fn from_previous_context(
+        previous_context: StateInitModeContext,
+        scope: &<DataFromJson as interactive_clap::ToInteractiveClapContextScope>::InteractiveClapContextScope,
+    ) -> color_eyre::eyre::Result<Self> {
+        let data = crate::common::parse_hex_kv_map(&scope.data)?;
+        Ok(Self(StateInitDataContext::build(
+            previous_context.code,
+            previous_context.global_context,
+            data,
+        )?))
+    }
+}
+
+#[derive(Debug, Clone, interactive_clap::InteractiveClap)]
+#[interactive_clap(input_context = StateInitDataContext)]
+#[interactive_clap(output_context = DepositContext)]
+pub struct Deposit {
+    /// How much do you want to deposit with the state init (e.g. '1 NEAR' or '0 NEAR')?
+    pub deposit: crate::types::near_token::NearToken,
+    #[interactive_clap(skip_default_input_arg)]
+    /// What is the signer account ID?
+    pub signer_account_id: crate::types::account_id::AccountId,
+    #[interactive_clap(named_arg)]
+    /// Select network
+    network_config: crate::network_for_transaction::NetworkForTransactionArgs,
+}
+
+impl Deposit {
+    pub fn input_signer_account_id(
+        context: &StateInitDataContext,
+    ) -> color_eyre::eyre::Result<Option<crate::types::account_id::AccountId>> {
+        crate::common::input_signer_account_id_from_used_account_list(
+            &context.global_context.config.credentials_home_dir,
+            "What is the signer account ID?",
+        )
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct DepositContext {
+    pub global_context: crate::GlobalContext,
+    pub state_init: near_primitives::deterministic_account_id::DeterministicAccountStateInit,
+    pub receiver_account_id: near_primitives::types::AccountId,
+    pub deposit: near_token::NearToken,
+    pub signer_account_id: near_primitives::types::AccountId,
+}
+
+impl DepositContext {
+    pub fn from_previous_context(
+        previous_context: StateInitDataContext,
+        scope: &<Deposit as interactive_clap::ToInteractiveClapContextScope>::InteractiveClapContextScope,
+    ) -> color_eyre::eyre::Result<Self> {
+        Ok(Self {
+            global_context: previous_context.global_context,
+            state_init: previous_context.state_init,
+            receiver_account_id: previous_context.receiver_account_id,
+            deposit: scope.deposit.into(),
+            signer_account_id: scope.signer_account_id.clone().into(),
+        })
+    }
+}
+
+impl From<DepositContext> for crate::commands::ActionContext {
+    fn from(item: DepositContext) -> Self {
+        let signer_id = item.signer_account_id.clone();
+        let receiver_id = item.receiver_account_id.clone();
+
+        let get_prepopulated_transaction_after_getting_network_callback: crate::commands::GetPrepopulatedTransactionAfterGettingNetworkCallback =
+            std::sync::Arc::new({
+                move |network_config| {
+                    use crate::common::JsonRpcClientExt as _;
+                    let receiver_id = &item.receiver_account_id;
+                    let result = network_config
+                        .json_rpc_client()
+                        .blocking_call_view_account(
+                            receiver_id,
+                            near_primitives::types::Finality::Final.into(),
+                        );
+                    // Best-effort check — only cancel if we positively confirm account exists.
+                    // All errors (UnknownAccount, network timeout, connection refused, etc.)
+                    // are treated as "proceed" to support the sign-later offline signing flow,
+                    // where the network may be unreachable at transaction construction time.
+                    if result.is_ok() {
+                        eprintln!(
+                            "\nDeterministic account <{}> already exists on <{}> network. No transaction needed.",
+                            receiver_id,
+                            network_config.network_name,
+                        );
+                        return Ok(crate::commands::PrepopulatedTransaction {
+                            signer_id: item.signer_account_id.clone(),
+                            receiver_id: item.receiver_account_id.clone(),
+                            actions: vec![],
+                        });
+                    }
+                    Ok(crate::commands::PrepopulatedTransaction {
+                        signer_id: item.signer_account_id.clone(),
+                        receiver_id: item.receiver_account_id.clone(),
+                        actions: vec![
+                            near_primitives::transaction::Action::DeterministicStateInit(Box::new(
+                                near_primitives::action::DeterministicStateInitAction {
+                                    state_init: item.state_init.clone(),
+                                    deposit: item.deposit,
+                                },
+                            )),
+                        ],
+                    })
+                }
+            });
+
+        Self {
+            global_context: item.global_context,
+            interacting_with_account_ids: vec![signer_id, receiver_id],
+            get_prepopulated_transaction_after_getting_network_callback,
+            on_before_signing_callback: std::sync::Arc::new(
+                |_prepopulated_unsigned_transaction, _network_config| Ok(()),
+            ),
+            on_before_sending_transaction_callback: std::sync::Arc::new(
+                |_signed_transaction, _network_config| Ok(String::new()),
+            ),
+            on_after_sending_transaction_callback: std::sync::Arc::new(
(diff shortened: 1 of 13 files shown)


src/common.rs
```diff
@@ -3133,3 +3151,23 @@ pub fn save_cli_command(cli_cmd_str: &str) {
...
+    input: &str,
+) -> color_eyre::eyre::Result<std::collections::BTreeMap<Vec<u8>, Vec<u8>>> {
+    let map: std::collections::BTreeMap<String, String> = serde_json::from_str(input)
+        .map_err(|e| color_eyre::eyre::eyre!("Failed to parse JSON: {e}"))?;
+    let mut result = std::collections::BTreeMap::new();
+    for (k, v) in map {
+        let key = hex::decode(&k)
+            .map_err(|e| color_eyre::eyre::eyre!("Failed to hex-decode key '{k}': {e}"))?;
+        let value = hex::decode(&v)
+            .map_err(|e| color_eyre::eyre::eyre!("Failed to hex-decode value '{v}': {e}"))?;
```
mitinarseny: I'd suggest to be consistent and encode as base64, as key-value pairs in `near_sdk::StateInit` and [`DeterministicAccountStateInitV1`](https://github.com/near/nearcore/blob/a8e277955a71fd6869fea96c7bc34d3fd2ea6c5f/core/primitives-core/src/deterministic_account_id.rs#L41-L43) are encoded as base64 as well

src/commands/contract/state_init/mod.rs
```diff
@@ -0,0 +1,332 @@
...
+        message = "data-from-file - Read hex-encoded key-value JSON data from a file"
+    ))]
+    /// Read hex-encoded key-value JSON data from a file
+    DataFromFile(DataFromFile),
+    #[strum_discriminants(strum(
+        message = "data-from-json - Provide hex-encoded key-value JSON data inline"
+    ))]
+    /// Provide hex-encoded key-value JSON data inline
+    DataFromJson(DataFromJson),
+}
```
mitinarseny, line 282: Maybe add an option to provide the whole [`StateInit`](https://github.com/near/near-sdk-rs/blob/34a244c1d67466ca5793d184f4075413f1d4d273/near-sdk/src/state_init.rs#L17-L19) serialized as borsh and encoded as base64 as well?

src/common.rs
pityjllk: hmmmm, tbh but since `state_init` in many cases is limited by ZBA and config usually gonna be just few fields like:
- account id
- hash
- number/counter
i think hex format is far more popular and easy to interpret than base64.... sure it uses base64 on the protocoal level but thats implementation detail right ... 🤔 also i like the other comment.  https://github.com/near/near-cli-rs/pull/560#discussion_r2863227832

so maybe we could have hex kv map, and option to pass raw base64 encoded state_init directly

src/commands/contract/state_init/mod.rs
```diff
@@ -0,0 +1,417 @@
...
+
+impl StateInitWithContractHashRef {
+    pub fn input_hash(
+        _context: &crate::GlobalContext,
+    ) -> color_eyre::eyre::Result<Option<crate::types::crypto_hash::CryptoHash>> {
+        Ok(Some(
+            inquire::CustomType::new("Enter the global contract code hash:").prompt()?,
+        ))
+    }
+}
```
vsavchyn-dev:
```suggestion
pub struct StateInitWithContractHashRef {
    /// Enter the global contract code hash:
    pub hash: crate::types::crypto_hash::CryptoHash,
    #[interactive_clap(subcommand)]
    data: Data,
}
```
Adding triple slash also shows a prompt (more of a nitpick but reduces the code size):

Image

src/commands/contract/state_init/mod.rs
```diff
@@ -0,0 +1,417 @@
...
+        }))
+    }
+}
+
+#[derive(Debug, Clone, interactive_clap::InteractiveClap)]
+#[interactive_clap(input_context = crate::GlobalContext)]
+#[interactive_clap(output_context = StateInitFromBorshBase64Context)]
+pub struct StateInitFromBorshBase64 {
+    #[interactive_clap(skip_default_input_arg)]
+    pub state_init_base64: String,
```
vsavchyn-dev: Maybe we can use `crate::types::Base64Bytes` for this field? Or am I missing something?

src/commands/contract/state_init/mod.rs
```diff
@@ -0,0 +1,417 @@
...
+
+impl StateInitFromBorshBase64 {
+    pub fn input_state_init_base64(
+        _context: &crate::GlobalContext,
+    ) -> color_eyre::eyre::Result<Option<String>> {
+        Ok(Some(
+            inquire::Text::new("Enter the borsh-base64 encoded StateInit:").prompt()?,
+        ))
+    }
+}
```
vsavchyn-dev: Making triple slash comment on `state_init_bas64` should enable removing this whole block of code

Unless there is a need for custom verification of the input (which in this case I believe might be nice here to parse and check the borsh-base64 string before going in-between contexts).

src/commands/contract/state_init/mod.rs
```diff
@@ -0,0 +1,417 @@
...
+    data: Data,
+}
+
+impl StateInitWithContractRefByAccount {
+    pub fn input_account_id(
+        _context: &crate::GlobalContext,
+    ) -> color_eyre::eyre::Result<Option<crate::types::account_id::AccountId>> {
+        Ok(Some(
+            inquire::CustomType::new("Enter the global contract account ID:").prompt()?,
+        ))
```
vsavchyn-dev:
```suggestion
        crate::common::input_non_signer_account_id_from_used_account_list(
            &context.global_context.config.credentials_home_dir,
            "Enter the global contract account ID:",
        )
```
How about we use `input_non_signer_account_id_from_used_account_list` here? Easier to input contract ids imo

src/commands/transaction/construct_transaction/state_init_receiver/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+
+impl StateInitWithContractHashRef {
+    pub fn input_hash(
+        _context: &super::ConstructTransactionSenderContext,
+    ) -> color_eyre::eyre::Result<Option<crate::types::crypto_hash::CryptoHash>> {
+        Ok(Some(
+            inquire::CustomType::new("Enter the global contract code hash:").prompt()?,
+        ))
+    }
+}
```
vsavchyn-dev:
```suggestion
pub struct StateInitWithContractHashRef {
    /// Enter the global contract code hash:
    pub hash: crate::types::crypto_hash::CryptoHash,
    #[interactive_clap(subcommand)]
    data: Data,
}
```

Same as step in `state-init` in `contract` sub-command

src/commands/transaction/construct_transaction/state_init_receiver/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+
+impl StateInitWithContractRefByAccount {
+    pub fn input_account_id(
+        _context: &super::ConstructTransactionSenderContext,
+    ) -> color_eyre::eyre::Result<Option<crate::types::account_id::AccountId>> {
+        Ok(Some(
+            inquire::CustomType::new("Enter the global contract account ID:").prompt()?,
+        ))
+    }
```
vsavchyn-dev:
```suggestion
        crate::common::input_signer_account_id_from_used_account_list(
            &context.global_context.config.credentials_home_dir,
            "Enter the global contract account ID: ",
        )
```
Same as step in state-init in contract sub-command

src/commands/contract/state_init/mod.rs
```diff
@@ -0,0 +1,417 @@
...
+        _context: &StateInitModeContext,
+    ) -> color_eyre::eyre::Result<Option<crate::types::path_buf::PathBuf>> {
+        Ok(Some(
+            inquire::CustomType::new(
+                "Enter the path to the file with base64-encoded key-value JSON data:",
+            )
+            .prompt()?,
+        ))
+    }
+}
```
vsavchyn-dev, line 292:
```suggestion
pub struct DataFromFile {
    /// Enter the path to the file with base64-encoded key-value JSON data:
    pub file_path: crate::types::path_buf::PathBuf,
    #[interactive_clap(named_arg)]
    deposit: Deposit,
}
```
Triple slash should also work here

src/commands/contract/state_init/mod.rs
```diff
@@ -0,0 +1,417 @@
...
+impl DataFromJson {
+    pub fn input_data(_context: &StateInitModeContext) -> color_eyre::eyre::Result<Option<String>> {
+        Ok(Some(
+            inquire::Text::new(
+                "Enter the base64-encoded key-value JSON data (e.g. '{\"AAEC\": \"AwQF\"}'):",
+            )
+            .prompt()?,
+        ))
+    }
+}
```
vsavchyn-dev, line 302:
```suggestion
pub struct DataFromJson {
    /// Enter the base64-encoded key-value JSON data (e.g. '{\"AAEC\": \"AwQF\"}'):
    pub data: String,
    #[interactive_clap(named_arg)]
    deposit: Deposit,
}
```
Same as previous

src/commands/contract/state_init/mod.rs
```diff
@@ -0,0 +1,417 @@
...
+            data,
+        )?))
+    }
+}
+
+#[derive(Debug, Clone, interactive_clap::InteractiveClap)]
+#[interactive_clap(input_context = StateInitDataContext)]
+#[interactive_clap(output_context = DepositContext)]
+pub struct Deposit {
+    pub deposit: crate::types::near_token::NearToken,
```
vsavchyn-dev, line 554: Let's add `input_deposit` function with `CustomType` prompting something like "What is the deposit for the state init call?"

Reference: https://github.com/near/near-cli-rs/blob/57565dea471954e153855cdc52b1c9a15ac5a1a4/src/transaction_signature_options/sign_with_mpc/mod.rs#L455-L479

src/commands/transaction/construct_transaction/state_init_receiver/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+
+impl StateInitFromBorshBase64 {
+    pub fn input_state_init_base64(
+        _context: &super::ConstructTransactionSenderContext,
+    ) -> color_eyre::eyre::Result<Option<String>> {
+        Ok(Some(
+            inquire::Text::new("Enter the borsh-base64 encoded StateInit:").prompt()?,
+        ))
+    }
+}
```
vsavchyn-dev:
```suggestion
pub struct StateInitFromBorshBase64 {
    /// Enter the borsh-base64 encoded StateInit
    pub state_init_base64: String,
    #[interactive_clap(named_arg)]
    deposit: Deposit,
}
```

maybe use `crate::types::Base64Bytes` here as well?

src/commands/transaction/construct_transaction/state_init_receiver/mod.rs
```diff
@@ -0,0 +1,356 @@
...
+            data,
+        )?))
+    }
+}
+
+#[derive(Debug, Clone, interactive_clap::InteractiveClap)]
+#[interactive_clap(input_context = StateInitDataContext)]
+#[interactive_clap(output_context = DepositContext)]
+pub struct Deposit {
+    pub deposit: crate::types::near_token::NearToken,
```
vsavchyn-dev, line 438: Let's add `input_deposit` function with `CustomType` prompting something like "What is the deposit for the state init call?"

Reference: https://github.com/near/near-cli-rs/blob/57565dea471954e153855cdc52b1c9a15ac5a1a4/src/transaction_signature_options/sign_with_mpc/mod.rs#L455-L479

vsavchyn-dev (changes requested):

src/commands/contract/state_init/mod.rs
pityjllk: addressed

src/commands/contract/state_init/mod.rs
pityjllk: addressed

src/commands/contract/state_init/mod.rs
```diff
@@ -0,0 +1,386 @@
...
+pub enum Data {
+    #[strum_discriminants(strum(
+        message = "data-from-file - Read base64-encoded key-value JSON data from a file"
+    ))]
+    DataFromFile(DataFromFile),
+    #[strum_discriminants(strum(
+        message = "data-from-json - Provide base64-encoded key-value JSON inline"
+    ))]
+    DataFromJson(DataFromJson),
+}
```
mitinarseny, line 282: I'd suggest to reorder them and improve the naming a bit:
```suggestion
pub enum Data {
    #[strum_discriminants(strum(
        message = "data-json - Provide base64-encoded key-value JSON inline"
    ))]
    DataJson(DataJson),
    #[strum_discriminants(strum(
        message = "data-json-file - Read base64-encoded key-value JSON data from a file"
    ))]
    DataJsonFile(DataJsonFile),
}
```

src/commands/contract/state_init/mod.rs
```diff
@@ -0,0 +1,386 @@
...
+    ))]
+    UseGlobalHash(StateInitWithContractHashRef),
+    #[strum_discriminants(strum(
+        message = "use-global-account-id - Use a global contract account ID"
+    ))]
+    UseGlobalAccountId(StateInitWithContractRefByAccount),
+    #[strum_discriminants(strum(
+        message = "from-borsh-base64     - Provide borsh serialized base64 encoded state init"
+    ))]
+    FromBorshBase64(StateInitFromBorshBase64),
```
mitinarseny, line 37: Also add an option for `from-borsh-file`?

mitinarseny (commented):
1. Can we also add support for adding more actions after StateInit? On it's own StateInit is not that useful, the most interesting stuff is when you combine it in one receipt 
```
near contract state-init ... [[add-action...] skip] sign-as ...
```
2. It would be nice if, in interactive mode, after `state-init use-global-account-id global.near data-json { ... }` I can see the derived `0s...` address, because further function calls might require to specify it in args

What do you think?

src/commands/contract/state_init/mod.rs
```diff
@@ -0,0 +1,386 @@
...
+}
+
+#[derive(Debug, Clone, interactive_clap::InteractiveClap)]
+#[interactive_clap(input_context = StateInitDataContext)]
+#[interactive_clap(output_context = DepositContext)]
+pub struct Deposit {
+    #[interactive_clap(skip_default_input_arg)]
+    pub deposit: crate::types::near_token::NearToken,
+    #[interactive_clap(skip_default_input_arg)]
+    pub signer_account_id: crate::types::account_id::AccountId,
```
mitinarseny: maybe add `sign-as` prefix?
```
state-init use-global-hash ... data-from-json {} deposit '0 NEAR' sign-as signer.near

**Review by vsavchyn-dev — APPROVED** [review]
LGTM! Thanks for contributing!

File: src/commands/contract/state_init/mod.rs (hunk 14 above)
Comment (vsavchyn-dev) [reply src/commands/contract/state_init/mod.rs:282]: I personally think that "data-from-file" and "data-from-json" sounds better, as in later steps we prompt for this data interactively.
Reordering on the other hand would be nice, as people might input raw json more

**Review by hlgltvnnk — APPROVED** [review]

**Comment by vsavchyn-dev** [comment]
Fyi it might make sense to make the gds util that you guys are creating a `near-cli` extension (it will pop up under extensions afaiu), as I believe that people usually would like to use near-cli in order to interact with NEAR blockchain, + having gds generate things via near-cli and not be tied to Cargo would also be nice.

I will get back to you with this idea once I fully am sure it can be done easily

P.S. I just simply believe that instead of this:
 ```
 near contract state-init \
  use-global-account-id 0s384bfa53f1718c7f53eaaa1b43c55e2aea3ef309 \
  data-from-json "$(cargo gds \
    --owner-id intents.sputnik-dao.near --index 42 \
    --approve 0x6c71114931fe91153b868f2cb29c5db70e59677d6d2e40404b3b9044d8052266 \
    --quiet)"
 ```
 
It would be better to have it as double-step action, with gds extension possibly generating near-cli command for quick run of the state-init action:
 ```
 near gds owner-id <owner.id> index <index> {no-approve or approve + <hash str>} {display or save-to-file + <path/to/file>}
 
 near contract state-init use-global-account-id \
   use-global-account-id 0s384bfa53f1718c7f53eaaa1b43c55e2aea3ef309
   data-from-file path/to/file
 ```

File: src/commands/contract/state_init/mod.rs [hunk 17]
```diff
@@ -0,0 +1,585 @@
...
+            .wrap_err("Failed to decode base64 from file content")?;
+        let state_init = crate::common::parse_borsh_base64_state_init(&data)?;
+        let receiver_account_id =
+            near_primitives::utils::derive_near_deterministic_account_id(&state_init);
+        Ok(Self(StateInitDataContext {
+            global_context: previous_context,
+            state_init,
+            receiver_account_id,
+        }))
+    }
```
mitinarseny: re-use `StateInitDataContext::build()`?

src/commands/contract/state_init/mod.rs
```diff
@@ -0,0 +1,585 @@
...
+pub struct PrintStateInitContext;
+
+impl PrintStateInitContext {
+    pub fn from_previous_context(
+        previous_context: StateInitDataContext,
+        _scope: &<PrintStateInit as interactive_clap::ToInteractiveClapContextScope>::InteractiveClapContextScope,
+    ) -> color_eyre::eyre::Result<Self> {
+        let bytes = borsh::to_vec(&previous_context.state_init)
+            .map_err(|e| color_eyre::eyre::eyre!("Failed to borsh-serialize state-init: {e}"))?;
+        println!("{}", near_primitives::serialize::to_base64(&bytes));
```
mitinarseny, line 485: do all other prints before print to stderr?

src/commands/contract/state_init/mod.rs
```diff
@@ -0,0 +1,585 @@
...
+            .map(|(k, v)| {
+                (
+                    near_primitives::serialize::to_base64(k),
+                    near_primitives::serialize::to_base64(v),
+                )
+            })
+            .collect();
+        println!(
+            "{}",
+            serde_json::to_string(&map).expect("BTreeMap<String, String> should be serializable")
```
mitinarseny: Maybe delegate it to serde_with?
```suggestion
struct Data(
    #[serde_as = "BTreeMap<Base64, Base64>"]
    BTreeMap<Vec<u8>, Vec<u8>>
);

let data = Data(data);
println!("{}", serde_json::to_string(&data).unwrap());
```

src/commands/contract/state_init/mod.rs
```diff
@@ -0,0 +1,585 @@
...
+                            "\nDeterministic account <{}> already exists on <{}> network. No transaction needed.",
+                            receiver_id,
+                            network_config.network_name,
+                        );
+                        return Ok(crate::commands::PrepopulatedTransaction {
+                            signer_id: item.signer_account_id.clone(),
+                            receiver_id: item.receiver_account_id.clone(),
+                            actions: vec![],
+                        });
+                    }
```
mitinarseny, line 672: It would be nice if this behavior can be configurable, i.e. force-include-state-init.
The reasons for that is:
1. I want to explicitly test for gas usage of overall tx on testnet, for example
1. This deterministic account can be deleted mid-flight by itself, so your new tx will fail

src/commands/transaction/construct_transaction/state_init_receiver/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+    ))]
+    UseGlobalAccountId(StateInitWithContractRefByAccount),
+    #[strum_discriminants(strum(
+        message = "from-borsh-base64      - Provide borsh serialized base64 encoded state init"
+    ))]
+    FromBorshBase64(StateInitFromBorshBase64),
+    #[strum_discriminants(strum(
+        message = "from-borsh-base64-file - Read borsh serialized base64 encoded state init from a file"
+    ))]
+    FromBorshBase64File(StateInitFromBorshBase64File),
```
mitinarseny: maybe require the file to contain borsh as-is, i.e. without additional base64 encoding?

src/common.rs
```diff
@@ -3184,3 +3202,34 @@ pub fn save_cli_command(cli_cmd_str: &str) {
...
+        .map_err(|e| color_eyre::eyre::eyre!("Failed to parse JSON: {e}"))?;
+    let mut result = std::collections::BTreeMap::new();
+    for (k, v) in map {
+        let key = near_primitives::serialize::from_base64(&k)
+            .map_err(|e| color_eyre::eyre::eyre!("Failed to base64-decode key '{k}': {e}"))?;
+        let value = near_primitives::serialize::from_base64(&v)
+            .map_err(|e| color_eyre::eyre::eyre!("Failed to base64-decode value '{v}': {e}"))?;
+        result.insert(key, value);
+    }
+    Ok(result)
```
mitinarseny: let serde_with handle that

src/commands/contract/state_init/mod.rs
pityjllk: and renamed to `new()`

src/commands/contract/state_init/mod.rs
pityjllk: sure 


```

cargo run -- contract state-init use-global-hash AE796LDMNCkKhudAdDAeU66bBT5mEFG7rdy4ErYyVKE3 data-from-json '{"": "100000006d61746465706c6f7965722e6e6561720909090909090909090909090909090909090909090909090909090909090909060606060606060606060607"}' print kv-map 2>/dev/null
{"":"100000006d61746465706c6f7965722e6e6561720909090909090909090909090909090909090909090909090909090909090909060606060606060606060607"}

```

src/commands/transaction/construct_transaction/state_init_receiver/mod.rs
pityjllk: that would be inconsistent with loading from file that is used for function arguments, so i would rather keep it aligned 🤔

src/commands/contract/state_init/mod.rs
pityjllk: nice

src/commands/contract/state_init/mod.rs
pityjllk: but its only for the state-init subcommand, since you cant attach any action to it(you can only do it through construct transaction) it doesnt make sense to proceed, becase if account exists state-init does nothing right ... 🤔

src/commands/transaction/construct_transaction/state_init_receiver/mod.rs
mitinarseny: AFAIU, it's the opposite: https://github.com/near/near-cli-rs/blob/7fc44334d9fea242893caf0445d8d14f318cf9eb/src/commands/contract/call_function/call_function_args_type/mod.rs#L104-L109

But then it makes sense to also add "save-to-file" option, which would write borsh-serialized state_init to file without base64 encoding it

src/commands/contract/state_init/mod.rs
```diff
@@ -0,0 +1,584 @@
...
+/// What would you like to do with the state-init?
+pub enum StateInitAction {
+    #[strum_discriminants(strum(
+        message = "send  - Submit a transaction to initialize the deterministic account"
+    ))]
+    /// Submit a transaction to initialize the deterministic account
+    Send(Deposit),
+    #[strum_discriminants(strum(message = "print - Print state-init details"))]
+    /// Print state-init details
+    Print(Print),
```
mitinarseny: Maybe rename `print` -> `inspect`, so that you can have `base64` and `save-to-file` options?
```suggestion
    #[strum_discriminants(strum(message = "inspect - Inspect state-init details"))]
    /// Inspect state-init details
    Inspect(Inspect),
```

src/commands/transaction/construct_transaction/state_init_receiver/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+}
+
+impl Deposit {
+    pub fn input_deposit(
+        context: &StateInitDataContext,
+    ) -> color_eyre::eyre::Result<Option<crate::types::near_token::NearToken>> {
+        eprintln!(
+            "\nDerived deterministic AccountId: {}",
+            context.receiver_account_id
+        );
```
mitinarseny: Can we come up with something that will preserve padding via `| ` on the left and dimmed font?

Image

src/common.rs
```diff
@@ -947,13 +953,26 @@ pub fn print_unsigned_transaction(
...
+                            for (i, (key, value)) in deterministic_account_state_init_v1.data.iter().enumerate() {
+                                let comma = if i < last_idx { "," } else { "" };
+                                ret.push_str(&format!(
+                                    "\n{:>45} \"{}\" : \"{}\"{}",
+                                    "", near_primitives::serialize::to_base64(key), near_primitives::serialize::to_base64(value), comma,
+                                ));
+                            }
+                            ret.push_str(&format!("\n{:>31} {:<13} }}", "", ""));
+                        }
                         ret.push_str(&format!("\n{:>31} {:<13} {}", "", "code", match deterministic_account_state_init_v1.code {
```
mitinarseny:
```suggestion
                            ret.push_str(&format!("\n{:>31} {:<13} {{}}", "", "data:"));
                        } else {
                            ret.push_str(&format!("\n{:>31} {:<13} {{", "", "data:"));
                            let last_idx = deterministic_account_state_init_v1.data.len() - 1;
                            for (i, (key, value)) in deterministic_account_state_init_v1.data.iter().enumerate() {
                                let comma = if i < last_idx { "," } else { "" };
                                ret.push_str(&format!(
                                    "\n{:>45} \"{}\" : \"{}\"{}",
                                    "", near_primitives::serialize::to_base64(key), near_primitives::serialize::to_base64(value), comma,
                                ));
                            }
                            ret.push_str(&format!("\n{:>31} {:<13} }}", "", ""));
                        }
ret.push_str(&format!("\n{:>31} {:<13} {}", "", "code:", match deterministic_account_state_init_v1.code {
```

mitinarseny (commented):
few more comments🙂

src/commands/transaction/construct_transaction/state_init_receiver/mod.rs
pityjllk: could not figure out a better way, tried including it into next prompt "What is the deposit..." but its even less readable... 

theoretically i could mimic asci chars and colors to try tom atch inquire exactly but i honestly thinkg its total waste of time but let me know if you want me to do that

src/commands/transaction/construct_transaction/mod.rs
```diff
@@ -51,13 +45,62 @@ impl ConstructTransaction {
...
     }
+}
+
+#[derive(Debug, EnumDiscriminants, Clone, interactive_clap::InteractiveClap)]
+#[interactive_clap(context = ConstructTransactionSenderContext)]
+#[strum_discriminants(derive(EnumMessage, EnumIter))]
+/// How would you like to specify the receiver?
+pub enum ReceiverMode {
+    #[strum_discriminants(strum(message = "account-id  - Specify receiver account ID directly"))]
+    AccountId(DirectReceiver),
```
mitinarseny: Maybe rename `account-id` -> `receiver-id`, so that it's more obvious from the cli one-liner what actually happens?
```suggestion
    #[strum_discriminants(strum(message = "receiver-id  - Specify receiver account ID directly"))]
    ReceiverId(DirectReceiver),
```

src/commands/transaction/construct_transaction/mod.rs
```diff
@@ -51,13 +45,62 @@ impl ConstructTransaction {
...
+    #[strum_discriminants(strum(
+        message = "state-init  - Derive receiver from deterministic state init (NEP-616)"
```
mitinarseny:
```suggestion
        message = "state-init  - Derive receiver ID from deterministic StateInit (NEP-616)"
```

src/commands/transaction/construct_transaction/state_init_receiver/mod.rs
mitinarseny: Okay, this indeed looks ugly 😂 The only option I found is via `.with_help_message()` for `deposit` but then this help message is removed as soon as you confirm the input, so it doesn't make sense to show it in the first place. Maybe then just get rid of this print? 🙃

src/commands/transaction/construct_transaction/state_init_receiver/mod.rs
```diff
@@ -0,0 +1,403 @@
...
+    #[interactive_clap(named_arg)]
+    deposit: Deposit,
+}
+
+impl DataFromJson {
+    pub fn input_data(_context: &StateInitModeContext) -> color_eyre::eyre::Result<Option<String>> {
+        Ok(Some(
+            inquire::Text::new(
+                "Enter the base64-encoded key-value JSON data (e.g. '{\"AAEC\": \"AwQF\"}'):",
+            )
```
mitinarseny, line 340: Maybe let's put examples in placeholders? They seem to look nicer in interactive mode
If so, then maybe add them all over the place? Like, for `use-global-hash`, `use-global-account-id`, `data-from-file`, etc...
```suggestion
            inquire::Text::new("Enter the base64-encoded key-value JSON data:")
                .with_placeholder(r#"{"AAEC": "AwQF"}"#)
```

mitinarseny (commented):
Looks AWESOME to me!
PS: left few final nits

src/commands/transaction/construct_transaction/state_init_receiver/mod.rs
vsavchyn-dev: I think that we can leave these examples as it is - we do this "e.g. ..." all around near-cli :)

src/commands/transaction/construct_transaction/state_init_receiver/mod.rs
frolvanya: FYI, you can use `tracing::info!(...)` to have a properly formatted message:
image

But overall, I think it's better to remove it entirely, because deterministic address is shown before sending transaction:
image

pityjllk: agreed, removed

src/commands/transaction/construct_transaction/state_init_receiver/mod.rs
pityjllk: then lets go for consistency and consider placeholers as separate improvment for future 🤔 ?

src/commands/transaction/construct_transaction/state_init_receiver/mod.rs
pityjllk: ok, changed to raw borsh

hlgltvnnk (approved):

src/commands/contract/state_init/mod.rs
```diff
@@ -0,0 +1,577 @@
...
+    ))]
+    UseGlobalAccountId(StateInitWithContractRefByAccount),
+    #[strum_discriminants(strum(
+        message = "from-borsh-base64      - Provide borsh serialized base64 encoded state init"
+    ))]
+    FromBorshBase64(StateInitFromBorshBase64),
+    #[strum_discriminants(strum(
+        message = "from-borsh-file        - Read borsh-serialized state init from a file"
+    ))]
+    FromBorshFile(StateInitFromBorshFile),
```
vsavchyn-dev, line 41:
```suggestion
    #[strum_discriminants(strum(
        message = "from-borsh-base64     - Provide borsh serialized base64 encoded state init"
    ))]
    FromBorshBase64(StateInitFromBorshBase64),
    #[strum_discriminants(strum(
        message = "from-borsh-file       - Read borsh-serialized state init from a file"
    ))]
    FromBorshFile(StateInitFromBorshFile),
```
To align these options with others

src/commands/contract/state_init/mod.rs
```diff
@@ -0,0 +1,577 @@
...
+pub enum StateInitAction {
+    #[strum_discriminants(strum(
+        message = "send  - Submit a transaction to initialize the deterministic account"
+    ))]
+    /// Submit a transaction to initialize the deterministic account
+    Send(Deposit),
+    #[strum_discriminants(strum(message = "inspect - Inspect state-init details"))]
+    /// Inspect state-init details
+    Inspect(Inspect),
+}
```
vsavchyn-dev, line 404: I am not sure what's the purpose of introducing the ability to inspect the state init action where you can check all things individually if we already display the unsigned transaction with state-init action after user selects network.

Image

Also, imo it's confusing to have two sends in CLI:
`near contract state-init ... data-from-json '{"AAEC": "AwQF" }' _send_ '0 NEAR' sign-as account.near network-config .... sign-with-keychain _send_`
I would imagine the flow to be similar to how `call-function as-transaction` works:
`near contract state-init ... data-from-json '{"AAEC": "AwQF" }' attached-deposit '0 NEAR' sign-as account.near network-config .... sign-with-keychain _send_`

Wdyt about removing inspect?

src/main.rs
```diff
@@ -102,26 +102,35 @@ fn main() -> crate::common::CliResult {
...
             }
             _ => {
-                match crate::js_command_match::JsCmd::try_parse() {
-                    Ok(js_cmd) => {
-                        let vec_cmd = js_cmd.rust_command_generation();
-                        let cmd = std::iter::once(near_cli_exec_path.to_owned()).chain(vec_cmd);
-                        Parser::parse_from(cmd)
+                // Backward compat: old `construct-transaction <sender> <receiver>` syntax
+                // Insert "account-id" subcommand before the bare receiver account ID.
+                if let Some(rewritten) = try_rewrite_construct_transaction_args() {
```
vsavchyn-dev, line 103: I am curious if this parsing of new `construct-transaction` will collide with near-cli-js commands that we try to parse?
From what I see in [near-cli-js](https://github.com/near/near-cli?tab=readme-ov-file#overview) it shouldn't, but maybe someone can spot an issue

src/commands/contract/state_init/mod.rs
vsavchyn-dev: Also, do we need `prepaid-gas` for this action? Just double-checking

vsavchyn-dev (commented):
From what I see it should work as intended, but I just have one question about `Inspect` for state-init - not sure if it should be there.

@frol - any other suggestions/remarks?

src/commands/contract/state_init/mod.rs
pityjllk: so the argument to actually have the inspect is for cli composability like scripts and be able to use deterministic account id as parameter in the function call

```
near-cli construct-transaction state-init <GLOBAL_CONTRACT_ID> <STATE_INIT> add-action function call <SOME_FUNCTION> with-json-args $(near-cli state-init <GLOBAL_CONTRACT_ID> <STATE_INIT> inspect deterministic-account-id))
```

vsavchyn-dev: I might be lacking context here, but why would I need to include the deterministic-account-id into function call args...? One of the examples could be memo for FT transfer in contract, right? What other usages we have for it?

~~Ultimately, I don't think that near-cli-rs was designed with composability in mind, considering we have some Y/N prompts in non-interactive mode (I might be wrong though, so correct me if I'm wrong). Also, I am not sure how many people use near-cli-rs in their scripts considering how easy it can fail on each step (would be nice though imo).~~

Plus, having two "sends" just doesn't sit right with me :)

EDIT: People actually use near-cli-rs in scripts.

mitinarseny: > I might be lacking context here, but why would I need to include the deterministic-account-id into function call args...? One of the examples could be memo for FT transfer in contract, right? What other usages we have for it?

This is a fundamental thing that is needed for sharded contract to work with each other as per [NEP-616](https://github.com/near/NEPs/blob/master/neps/nep-0616.md). As an example, consider transferring [sharded FT](https://github.com/mitinarseny/near-sdk-rs/blob/3c3b9b498df3fb4c0c834fa869b369b1fbc16cb6/examples/sharded-fungible-token/wallet/src/lib.rs#L178-L191) to not-yet-created receiver. You'd need to derive its AccountId and attach StateInit to the original `sft_send()` function call.

pityjllk: fyi: renamed 'send' to 'execute'

pityjllk:
i think the argument was that since borsh is natvie serialization used at the protocol level then it feels right to support it, it might be useful for debugging etc. also effort has been made already so i would keep it ....

src/commands/contract/state_init/mod.rs
```diff
@@ -0,0 +1,618 @@
...
+    pub fn from_previous_context(
+        previous_context: StateInitDataContext,
+        _scope: &<InspectStateInitJson as interactive_clap::ToInteractiveClapContextScope>::InteractiveClapContextScope,
+    ) -> color_eyre::eyre::Result<Self> {
+        println!(
+            "{}",
+            serde_json::to_string_pretty(&previous_context.state_init).map_err(
+                |e| color_eyre::eyre::eyre!("Failed to serialize state-init to JSON: {e}")
+            )?
+        );
```
mitinarseny, line 509: Maybe only prettify if stdout is tty?

src/commands/contract/state_init/mod.rs
```diff
@@ -0,0 +1,618 @@
...
+    ))]
+    UseGlobalAccountId(StateInitWithContractRefByAccount),
+    #[strum_discriminants(strum(
+        message = "from-borsh-base64     - Provide borsh serialized base64 encoded state init"
+    ))]
+    FromBorshBase64(StateInitFromBorshBase64),
+    #[strum_discriminants(strum(
+        message = "from-borsh-file       - Read borsh-serialized state init from a file"
+    ))]
+    FromBorshFile(StateInitFromBorshFile),
```
mitinarseny, line 41: Maybe also add `from-json[-file]` for the sake of roundtrip with `near contract state-init ... inspect json`?
And same for `near transaction construct-transaction receiver.near state-init from-json[-file]`?

mitinarseny (commented):
few more comments on JSON part

src/commands/contract/state_init/mod.rs
pityjllk: actually in the whole repo `to_string_pretty` is used, so it would be insconsisted, so ideally that could be addressed in separate pr?

src/commands/contract/state_init/mod.rs
```diff
@@ -0,0 +1,700 @@
...
+    #[strum_discriminants(strum(
+        message = "from-json             - Provide JSON-serialized state init inline"
+    ))]
+    FromJson(StateInitFromJson),
+    #[strum_discriminants(strum(
+        message = "from-json-file        - Read JSON-serialized state init from a file"
+    ))]
+    FromJsonFile(StateInitFromJsonFile),
```
mitinarseny: Maybe switch places and put json-related first, since they would be more frequently used in the interactive mode?

src/commands/contract/state_init/mod.rs
```diff
@@ -0,0 +1,700 @@
...
+#[interactive_clap(context = StateInitDataContext)]
+#[strum_discriminants(derive(EnumMessage, EnumIter))]
+/// In which format would you like to display the state-init?
+pub enum InspectStateInitFormat {
+    #[strum_discriminants(strum(message = "borsh - Borsh-serialized base64"))]
+    /// Borsh-serialized base64
+    Borsh(InspectStateInitBorsh),
+    #[strum_discriminants(strum(message = "json  - JSON-serialized"))]
+    /// JSON-serialized
+    Json(InspectStateInitJson),
```
mitinarseny: Also here: put json first?

mitinarseny (commented):
I hope these are final nits from me, sorry 🙃

mitinarseny:
Note: before merging it we need to upgrade `near-primitives-core` to a version which includes https://github.com/near/nearcore/pull/15539

src/commands/transaction/construct_transaction/state_init_receiver/mod.rs
```diff
@@ -0,0 +1,482 @@
...
+    ))]
+    FromBorshFile(StateInitFromBorshFile),
+    #[strum_discriminants(strum(
+        message = "from-json              - Provide JSON-serialized state init inline"
+    ))]
+    FromJson(StateInitFromJson),
+    #[strum_discriminants(strum(
+        message = "from-json-file         - Read JSON-serialized state init from a file"
+    ))]
+    FromJsonFile(StateInitFromJsonFile),
```
mitinarseny: Put JSON options first here as well?
```suggestion
    #[strum_discriminants(strum(
        message = "from-json              - Provide JSON-serialized state init inline"
    ))]
    FromJson(StateInitFromJson),
    #[strum_discriminants(strum(
        message = "from-json-file         - Read JSON-serialized state init from a file"
    ))]
    FromJsonFile(StateInitFromJsonFile),
    #[strum_discriminants(strum(
        message = "from-borsh-base64      - Provide borsh serialized base64 encoded state init"
    ))]
    FromBorshBase64(StateInitFromBorshBase64),
    #[strum_discriminants(strum(
        message = "from-borsh-file        - Read borsh-serialized state init from a file"
    ))]
    FromBorshFile(StateInitFromBorshFile),
```

src/common.rs
```diff
@@ -951,13 +957,26 @@ pub fn print_unsigned_transaction(
...
+                            ret.push_str(&format!("\n{:>31} {:<13} }}", "", ""));
+                        }
+                        ret.push_str(&format!("\n{:>31} {:<13} {}", "", "code:", match deterministic_account_state_init_v1.code {
                             GlobalContractIdentifier::CodeHash(hash) => {
-                                format!("use global <{hash}> code to deploy from")
+                                format!("hash: {hash}")
                             }
                             GlobalContractIdentifier::AccountId(ref account_id) => {
-                                format!("use global <{account_id}> code to deploy from")
+                                format!("account ref: {account_id}")
```
mitinarseny: Since we've standardized JSON representation of `StateInit`, then maybe print it as a JSON instead of writing custom human-readable serialization?

pityjllk:
```
│       -- create deterministic account <0s2de1179d14651ce32c16bca892d2e804057b6fcb>:
│                       deposit     : 0 NEAR
│                       state-init  : {
│                                       "V1": {
│                                         "code": {
│                                           "AccountId": "0s29e346108955b88c2d180a4ba17662b1f2cc1028"
│                                         },
│                                         "data": {
│                                           "": "GAAAAGludGVudHMuc3B1dG5pay1kYW8ubmVhcgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAs4NZ2fFCre1avLgVH/oTQj8p5Mx4pXhGj5Xov8nIgpDI="
│                                         }
│                                       }
│                                     }
│

```

pityjllk: how about printing whole state init as json?

mitinarseny: Yes, that's exactly what I was meant, nice!

mitinarseny (approved):
Thanks @pityjllk, LGTM!

vsavchyn-dev (approved):
Stamp, thanks for the new command and hard work!

race-of-sloths:
@pityjllk Thank you for your contribution! Your pull request is now a part of the [Race of Sloths](https://race-of-sloths.com)!
Do you want to apply for monthly streak? Get 8+ score for a single PR this month and receive boost for race-of-sloths!

[<picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://race-of-sloths.com/images/bot-banner-dark.svg">
    <source media="(prefers-color-scheme: light)" srcset="https://race-of-sloths.com/images/bot-banner-light.svg">
    Shows inviting banner with latest news.
</picture>](https://race-of-sloths.com/latest-news)

[<picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://badge.race-of-sloths.com/pityjllk?type=bot&pr=near/near-cli-rs/560&theme=dark">
    <source media="(prefers-color-scheme: light)" srcset="https://badge.race-of-sloths.com/pityjllk?type=bot&pr=near/near-cli-rs/560&theme=light">
    Shows profile picture for the author of the PR
</picture>](https://race-of-sloths.com/profile/pityjllk)




<details><summary>Current status: <i>executed</i></summary>

| Reviewer | Score |
|--------|--------|
| @vsavchyn-dev  | 13 |


Your contribution is much appreciated with a final score of 13!
You have received 140 (130 base + 10 monthly bonus) Sloth points for this contribution

@vsavchyn-dev received 25 Sloth Points for reviewing and scoring this pull request.

Congratulations @pityjllk! Your PR was highly scored and you completed another monthly streak! To keep your monthly streak make another pull request next month and get 8+ score for it
</details><details><summary>What is the Race of Sloths</summary>
<p>
Race of Sloths is a friendly competition where you can participate in challenges and compete with other open-source contributors within your normal workflow

For contributors:

- Tag @race-of-sloths inside your pull requests
- Wait for the maintainer to review and score your pull request
- Check out your position in the [Leaderboard](https://race-of-sloths.com/leaderboard)
- Keep weekly and monthly streaks to reach higher positions
- Boast your contributions with a dynamic picture of your [Profile](https://race-of-sloths.com/profile/pityjllk)

For maintainers:
- Score pull requests that participate in the Race of Sloths and receive a reward
- Engage contributors with fair scoring and fast responses so they keep their streaks
- Promote the Race to the point where the Race starts promoting you
- Grow the community of your contributors

Feel free to check [our website](https://race-of-sloths.com) for additional details!
</p>
</details> 

<details>
<summary>Bot commands</summary>

- For contributors
  - **Include a PR:** `@race-of-sloths include` to enter the Race with your PR
- For maintainers:
  - **Invite contributor** `@race-of-sloths invite` to invite the contributor to participate in a race or include it, if it's already a runner.
  - **Assign points:** `@race-of-sloths score [1/2/3/5/8/13]` to award points based on your assessment.
  - **Reject this PR:** `@race-of-sloths exclude` to send this PR back to the drawing board.
  - **Exclude repo:** `@race-of-sloths pause` to stop bot activity in this repo until `@race-of-sloths unpause` command is called

</details>

vsavchyn-dev:
@race-of-sloths score 13

PR #564, opened 2026-03-05
fix: wait for block finality before returning from send

r-near:
## Summary

- Switch the send path from `broadcast_tx_commit` to `send_tx` with `wait_until: TxExecutionStatus::Final`
- When the CLI reports a transaction as successful, the result is now visible to subsequent `Finality::Final` queries

## Problem

`broadcast_tx_commit` returns once the transaction is **executed**, but before the block reaches **doomslug finality**. View commands like `near account view-account-summary ... now` query with `Finality::Final`, which reads the latest finalized block — typically 2-3 blocks behind the chain head.

This causes a race condition in scripts where a `create-account` succeeds but an immediately following `view-account-summary` fails with "account does not exist" because the finalized block hasn't caught up yet.

Reported in https://nearone.slack.com/archives/C07UW93JVQ8/p1772634083567219

## Test plan

- [ ] Verify `cargo check` passes (confirmed locally)
- [ ] Test that `near account create-account ... send` followed by `near account view-account-summary ... now` works without a sleep on localnet

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -3175,20 +3175,7 @@ fn get_used_ft_contract_account_list_path(
 }
 
 pub fn is_used_ft_contract_account_list_exist(credentials_home_dir: &std::path::Path) -> bool {
-    let ft_contract_account_list_path =
-        get_used_ft_contract_account_list_path(credentials_home_dir);
-    if !ft_contract_account_list_path.exists() {
-        return false;
-    }
-
-    match std::fs::read_to_string(&ft_contract_account_list_path) {
-        Ok(content) => {
-            serde_json::from_str::<VecDeque<crate::types::ft_properties::FtContract>>(&content)
-                .map(|list| !list.is_empty())
-                .unwrap_or(false)
-        }
-        Err(_) => false,
-    }
+    get_used_ft_contract_account_list_path(credentials_home_dir).exists()
 }
 
 fn get_top_ft_tokens_from_nearblocks()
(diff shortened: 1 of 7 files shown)


src/common.rs
```diff
@@ -2736,7 +2736,8 @@ fn check_request_payload_for_broadcast_tx_commit(
     let method = request_payload.get("method").cloned();
     let params_value = request_payload.get("params").cloned();
     if let Some(method) = method
-        && method.to_string().contains("broadcast_tx_commit")
+        && (method.to_string().contains("broadcast_tx_commit")
+            || method.to_string().contains("send_tx"))
```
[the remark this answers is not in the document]
r-near: [`42e0720` (this PR)](https://github.com/near/near-cli-rs/pull/564/commits/42e07200af92d3f8238bdf0b9393cee42963e2b6)

src/transaction_signature_options/send/mod.rs
```diff
@@ -102,15 +102,15 @@ pub fn sending_signed_transaction(
     let retries_number = 5;
     let mut retries = (1..=retries_number).rev();
     let transaction_info = loop {
-        let request =
-            near_jsonrpc_client::methods::broadcast_tx_commit::RpcBroadcastTxCommitRequest {
-                signed_transaction: signed_transaction.clone(),
-            };
+        let request = near_jsonrpc_client::methods::send_tx::RpcSendTransactionRequest {
+            signed_transaction: signed_transaction.clone(),
+            wait_until: near_primitives::views::TxExecutionStatus::Final,
```
frolvanya: Maybe it's better to make finality configurable across the near-cli with optional `finality` field which would default to `Final`? I often send transactions to mpc and they can take a while to execute, so I have to interrupt cli process to prevent possible retries (and overall unblock terminal). It'd be nice if I could just set my custom finality to `Included`

r-near: Good idea — opened #565 to track this separately.

PR #568, opened 2026-03-07
feat: allow custom input as method name

frolvanya:
In current approach methods are filtered out based method kind (view/change) and we can't pass custom method name, so I've changed from Selection to Text + with_autocomplete approach 

image
image

### DIFF
--- a/src/commands/contract/call_function/mod.rs
+++ b/src/commands/contract/call_function/mod.rs
@@ -1,4 +1,5 @@
 use inquire::Text;
+use near_abi::AbiFunctionKind;
 use strum::{EnumDiscriminants, EnumIter, EnumMessage};
 
 mod as_read_only;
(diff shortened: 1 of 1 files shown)


src/commands/contract/call_function/mod.rs
```diff
@@ -71,22 +73,25 @@ fn input_function_name(
                     contract_account_id,
                 ))
         {
-            let function_names = contract_abi
+            function_names = contract_abi
                 .body
                 .functions
                 .into_iter()
                 .filter(|function| function_kind == function.kind)
```
frolvanya:
```suggestion
                .filter(|function| {
                    function.kind == AbiFunctionKind::View || function_kind == function.kind
                })
```

Also, it might be a good idea to allow calling view method in change calls, since it's allowed. But I don't have a strong preference here, so let me know what will be better

frol: Yeah, let's allow that as well

frolvanya: Added in ee42ae49900c3b989c40c572191504b207d9e5b1

PR #574, opened 2026-03-14
feat: Improved interactive contract selection option for FT tokens

### LINKED ISSUE #567: [Task] Simplify interaction with tokens
### Background

Right now users have to specify full token address to interact with FT, while near's account id make it easier in most of the cases and some tokens like `wrap.near` are easy to remember, other popular tokens like USDC are using explicit accounts (17208628f84f5d6ad33f0da3bbbeb27ffcb398eac501a31bd6ad2011e36133a1), so users have to open nearblocks or any other explorer to copy token address first. In my opinion, this UX can be improved. For example, we could store popular tokens or tokens that user recently interacted with in `~/.near-credentials/tokens.json` and give them to user when needed in a human readable format. Example of usage I'd like to see:
```bash
$ near tokens frolik.near send-ft
◆  What is the ft-contract account ID?  
│   ○ USDC (17208628f84f5d6ad33f0da3bbbeb27ffcb398eac501a31bd6ad2011e36133a1)
│   ○ wNEAR (wrap.near)
│   ○ wNEAR (wrap.testnet)
│   ○ Tether USD (usdt.tether-token.near)
│   ○ ZEC (zec.omft.near)
│   ○ Ether (eth.bridge.near)
│  ↓○ Aurora (aaaaaa20d9e0e2461697782ef11675f668207961.factory.bridge.near)
```

> [!NOTE]
> User still should be able to pass custom token contract id, it must not be a "selection-only" approach like in contract method calling implementation (this part should be also fixed IMO, because ABI marks methods with `&self` as `view` and it's impossible to select method in this case. But it can be solved in a separate PR (#568)

### User Story

As a user, I'd like to have a simple token interaction with popular or recent tokens

### Acceptance Criteria

Recent and popular tokens should be stored somewhere and given to user to pick whenever it's needed. See my vision of how I'd like to see interaction in `Background` section

### Resources & Additional Notes

This is also a good first step for implementing intents swap extension for near-cli-rs

FroVolod:
Resolves #567 

Interactive mode:

https://github.com/user-attachments/assets/4d22cc5d-04d4-45a0-a0ab-464bea14b0fa

Command line mode:

https://github.com/user-attachments/assets/abe6ac1d-01f7-47e9-8839-e36d0afffc2b

ft_contracts.json:
Screenshot 2026-03-14 at 12 10 37

### DIFF
--- a/src/commands/tokens/send_ft/mod.rs
+++ b/src/commands/tokens/send_ft/mod.rs
@@ -37,24 +37,22 @@ impl FtContractContext {
         let ft_contract_account_id: near_primitives::types::AccountId =
             scope.ft_contract_account_id.clone().into();
 
-        let ft_metadata = {
-            let network_config = crate::common::find_network_where_account_exist(
-                &previous_context.global_context,
-                ft_contract_account_id.clone(),
-            )?
-            .wrap_err_with(|| {
-                format!(
-                    "Contract <{}> does not exist in networks",
-                    ft_contract_account_id
-                )
-            })?;
+        let network_config = crate::common::find_network_where_account_exist(
+            &previous_context.global_context,
+            ft_contract_account_id.clone(),
+        )?
+        .wrap_err_with(|| {
+            format!(
+                "Contract <{}> does not exist in networks",
+                ft_contract_account_id
+            )
+        })?;
 
-            crate::types::ft_properties::params_ft_metadata(
-                ft_contract_account_id.clone(),
-                &network_config,
-                near_primitives::types::Finality::Final.into(),
-            )?
-        };
+        let ft_metadata = crate::types::ft_properties::params_ft_metadata(
+            ft_contract_account_id.clone(),
+            &network_config,
+            near_primitives::types::Finality::Final.into(),
+        )?;
 
         let ft_contract = crate::types::ft_properties::FtContract {
             ft_metadata,


src/common.rs
```diff
@@ -3167,6 +3167,72 @@ fn input_account_id_from_used_account_list(
...
+                == new_ft_contract_account.ft_contract_account_id
+        })
+        .and_then(|position| ft_contract_accounts_list.remove(position))
+    {
+        used_ft_contract_account
+    } else {
+        new_ft_contract_account.clone()
+    };
+
+    ft_contract_accounts_list.push_front(used_ft_contract_account);
```
frolvanya, line 3302: I think we should cap the list length to some constant (around 100 or even less)

```suggestion
    ft_contract_accounts_list.push_front(used_ft_contract_account);
    ft_contract_accounts_list.truncate(FT_CONTRACT_ACCOUNTS_LIST_SIZE_CAP);
```

FroVolod: done ([e11593e](https://github.com/near/near-cli-rs/pull/574/commits/e11593e3ec99fb8ae352ca6f7a2ad3e140c18588))

src/common.rs
```diff
@@ -3174,21 +3174,98 @@ fn get_used_ft_contract_account_list_path(
...
+        Err(_) => false,
+    }
 }
 
-pub fn create_used_ft_contract_account_list(
-    credentials_home_dir: &std::path::Path,
-) -> color_eyre::eyre::Result<()> {
+fn get_top_ft_tokens_from_nearblocks() -> Result<
+    VecDeque<crate::types::ft_properties::FtContract>,
+    Box<dyn std::error::Error + Send + Sync>,
```
frolvanya: nitpick: why not `color_eyre::Result` or `Result<VecDeque<...>, String>`?

src/commands/tokens/send_ft/mod.rs
```diff
@@ -10,11 +12,140 @@ mod amount_ft;
...
+        .map(|ft_contract| {
+            (
+                ft_contract.ft_contract_account_id.clone(),
+                ft_contract.ft_metadata.clone(),
+            )
+        })
+        .collect();
+
+        let ft_metadata = if let Some(ft_metadata) = hash_map.get(&ft_contract_account_id) {
+            ft_metadata.clone()
```
frolvanya: I know that in reality this should never happen, but for example, user can fetch top 100 tokens now and then token owner can change decimals in a while. In this case our amount normalization can fail (user asked to transfer 10 XYZ, old decimals were set to 6, not it's set to 8, so user accidentally transfers 1000 XYZ). I guess small latency optimization is not worth the risk and it's better to always fetch live metadata

src/common.rs
```diff
@@ -3167,6 +3167,149 @@ fn input_account_id_from_used_account_list(
...
+> {
+    #[derive(serde::Deserialize)]
+    struct ApiResponse {
+        tokens: VecDeque<crate::types::ft_properties::FtContract>,
+    }
+
+    let url = url::Url::parse("https://api.nearblocks.io/v1/fts")?;
```
frolvanya: informational: this fetches top 50 tokens instead of top 100 due to nearblocks pagination. I was thinking about top 100, but checked these tokens on nearblocks and mostly they're not that popular, so I think we can leave it as it is

src/common.rs
```diff
@@ -3174,21 +3174,98 @@ fn get_used_ft_contract_account_list_path(
...
-            "Failed to create file: {:?}",
-            &ft_contract_account_list_path
-        )
-    })?;
+
+    if let Ok(ft_contract_accounts_list) = get_top_ft_tokens_from_nearblocks() {
+        let ft_contract_account_list_buf = serde_json::to_string(
+            &ft_contract_accounts_list
+                .into_iter()
+                .take(FT_CONTRACT_ACCOUNTS_LIST_SIZE_CAP)
```
frolvanya, line 3248: I believe this is never used, right? Because nearblocks always returns up to 50 top tokens
https://api.nearblocks.io/api-docs#tag/fts/GET/v1/fts

Image

src/common.rs
frolvanya: I think not entirely, because new tokens still can be added and it will result to >50 tokens stored in file or did I miss something?

Related to: https://github.com/near/near-cli-rs/pull/574#discussion_r2950134514

frolvanya:
> @frolvanya 
> 
> > Also, that's more like a topic for a discussion rather than my request, but is it possible to separate tokens by network?
> 
> 
> 
> I basically did not add ft-contracts from https://api-testnet.nearblocks.io, because I believe that the "testnet" user will be able to find the ft-contract he needs on his own. In addition, this contract will be saved in his file ("ft_contracts.json") for further work.
> 
> If my opinion is wrong, please correct me.
> 
> 
> 
> 

I'm not talking about fetching top testnet tokens, because it would be useless indeed. I think people are using like 2-3 tokens like wnear or usdc and some custom tokens at most. I'm talking about not showing mainnet tokens if user wrote that his address is alice.testnet and showing only recent testnet tokens

P.S: but again, not sure if we need to over complicate things, maybe it's better to leave it as it's right now

src/common.rs
FroVolod: fixed ([98402fc](https://github.com/near/near-cli-rs/pull/574/commits/98402fc2b95a87210338e8323f823b1101f44c2d))

src/common.rs
FroVolod: > I think we should cap the list length to some constant (around 100 or even less)

I believe user-added ft-contracts would fit into your proposal.

src/commands/tokens/send_ft/mod.rs
FroVolod: I believe this is theoretically possible only on a "testnet" since the contracts are immutable.

src/common.rs
FroVolod: I thought so too when I introduced the restrictions.

src/common.rs
FroVolod: This is a precaution in case nearblocks changes the default value.

src/commands/tokens/send_ft/mod.rs
frolvanya: > I believe this is theoretically possible only on a "testnet" since the contracts are immutable.

Not always, for example, we can update metadata (including decimals) with DAO. Sure, we won't update decimals, but I believe we shouldn't risk it for a minor latency gain

https://github.com/Near-One/omni-bridge/blob/9b172eac9c1144aa246458505cdc19bd5f190272/near/omni-token/src/lib.rs#L181

FroVolod: done ([9af4634](https://github.com/near/near-cli-rs/pull/574/commits/9af4634648eb3fd6b0baf67f35d8c5e7848902cc))

src/commands/tokens/send_ft/mod.rs
```diff
@@ -10,11 +12,129 @@ mod amount_ft;
...
+                ft_contract_account_id.clone(),
+            )?
+            .wrap_err_with(|| {
+                format!(
+                    "Contract <{}> does not exist in networks",
+                    ft_contract_account_id
+                )
+            })?;
+
+            crate::types::ft_properties::params_ft_metadata(
```
frolvanya: nitpick: this scope is not needed, you can move network config outside and set `let ft_metdata = crate::types::ft_properties::params_ft_metadata(...)` directly without making a scope for that

src/main.rs
```diff
@@ -136,6 +132,16 @@ fn main() -> crate::common::CliResult {
     };
     near_cli_rs::setup_tracing(verbosity)?;
 
+    if !crate::common::is_used_account_list_exist(&config.credentials_home_dir) {
+        let _ = crate::common::create_used_account_list_from_legacy_keychain(
```
frolvanya, line 136: I'd like to see some sort of warning there, not just silence result completely. Will it look good if we just using `tracing::warn!` or it should be done differently?

src/main.rs
```diff
@@ -136,6 +132,16 @@ fn main() -> crate::common::CliResult {
...
+            &config.credentials_home_dir,
+        );
+    }
+
+    if !crate::common::is_used_ft_contract_account_list_exist(&config.credentials_home_dir) {
+        let _ = crate::common::create_used_ft_contract_account_list(&config.credentials_home_dir);
```
frolvanya, line 142: same here

src/main.rs
FroVolod: Provided explanation in the example for ft_contract.json

src/main.rs
FroVolod: I also initially thought of displaying a warning, but this text is confusing to the user. I decided that it would be better without a warning.
Screenshot 2026-03-20 at 18 33 55

src/commands/tokens/send_ft/mod.rs
FroVolod: done ([c05bfe8](https://github.com/near/near-cli-rs/pull/574/commits/c05bfe8335e3ade8da7a61d1c4b330f434fc0ea2))

PR #582, opened 2026-04-02
feat: validate receiver account before sending tokens

Gmin2:
Adds recipient validation to the tokens subcommand (send-near, send-ft,send-ft-call, send-nft), after network selection, two checks run: a network mismatch check that warns if the receiver looks like it belongs to a different network  and an rpc existence check that warns if the receiver account doesn't exist on the selected network. Both checks prompt the user for confirmation and never block, the user can always choose to proceed. If the rpc is unreachable,the existence check is skipped silently.

image

Fixes #581

### DIFF
--- a/src/commands/tokens/send_ft/amount_ft.rs
+++ b/src/commands/tokens/send_ft/amount_ft.rs
@@ -109,6 +109,17 @@ impl FtTransferParamsContext {
                 let deposit = scope.deposit.unwrap_or(crate::types::near_token::NearToken::from_yoctonear(1));
 
                 move |network_config| {
+                    if !crate::common::validate_receiver_account_id(
+                        network_config,
+                        &receiver_account_id,
+                    )? {
+                        return Ok(crate::commands::PrepopulatedTransaction {
+                            signer_id: signer_account_id.clone(),
+                            receiver_id: ft_contract_account_id.clone(),
+                            actions: vec![],
+                        });
+                    }
+
                     let amount_ft = if let crate::types::ft_properties::FungibleTokenTransferAmount::ExactAmount(ft) = &ft_transfer_amount {
                         ft
                     } else {
--- a/src/commands/tokens/send_ft/amount_ft.rs
+++ b/src/commands/tokens/send_ft/amount_ft.rs
@@ -114,6 +114,7 @@ impl FtTransferParamsContext {
                         network_config,
                         &receiver_account_id,
                         verbosity,
+                        previous_context.global_context.offline,
                     )? {
                         return Ok(crate::commands::PrepopulatedTransaction {
                             signer_id: signer_account_id.clone(),
--- a/src/commands/tokens/send_ft/amount_ft.rs
+++ b/src/commands/tokens/send_ft/amount_ft.rs
@@ -107,11 +107,13 @@ impl FtTransferParamsContext {
                 let memo = scope.memo.trim().to_string();
                 let gas = scope.gas.unwrap_or(near_gas::NearGas::from_tgas(100));
                 let deposit = scope.deposit.unwrap_or(crate::types::near_token::NearToken::from_yoctonear(1));
+                let verbosity = previous_context.global_context.verbosity;
 
                 move |network_config| {
                     if !crate::common::validate_receiver_account_id(
                         network_config,
                         &receiver_account_id,
+                        verbosity,
                     )? {
                         return Ok(crate::commands::PrepopulatedTransaction {
                             signer_id: signer_account_id.clone(),
--- a/src/commands/tokens/send_ft_call/amount_ft.rs
+++ b/src/commands/tokens/send_ft_call/amount_ft.rs
@@ -224,6 +224,7 @@ fn build_action_context(
                     network_config,
                     &receiver_account_id,
                     verbosity,
+                    previous_context.global_context.offline,
                 )? {
                     return Ok(crate::commands::PrepopulatedTransaction {
                         signer_id: signer_account_id.clone(),
--- a/src/commands/tokens/send_ft_call/amount_ft.rs
+++ b/src/commands/tokens/send_ft_call/amount_ft.rs
@@ -217,11 +217,13 @@ fn build_action_context(
             let msg = msg.clone();
             let gas = previous_context.gas;
             let deposit = previous_context.deposit;
+            let verbosity = previous_context.global_context.verbosity;
 
             move |network_config| {
                 if !crate::common::validate_receiver_account_id(
                     network_config,
                     &receiver_account_id,
+                    verbosity,
                 )? {
                     return Ok(crate::commands::PrepopulatedTransaction {
                         signer_id: signer_account_id.clone(),
--- a/src/commands/tokens/send_ft_call/amount_ft.rs
+++ b/src/commands/tokens/send_ft_call/amount_ft.rs
@@ -219,6 +219,17 @@ fn build_action_context(
             let deposit = previous_context.deposit;
 
             move |network_config| {
+                if !crate::common::validate_receiver_account_id(
+                    network_config,
+                    &receiver_account_id,
+                )? {
+                    return Ok(crate::commands::PrepopulatedTransaction {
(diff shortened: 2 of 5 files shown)


src/commands/tokens/send_ft/amount_ft.rs
```diff
@@ -109,6 +109,17 @@ impl FtTransferParamsContext {
...
+                        network_config,
+                        &receiver_account_id,
+                    )? {
+                        return Ok(crate::commands::PrepopulatedTransaction {
+                            signer_id: signer_account_id.clone(),
+                            receiver_id: ft_contract_account_id.clone(),
+                            actions: vec![],
+                        });
+                    }
+
```
FroVolod, line 125: You cannot use the user dialog in `fn from previous context()`. Some users use a script command that should end with "quiet".

src/commands/tokens/send_ft_call/amount_ft.rs
```diff
@@ -219,6 +219,17 @@ fn build_action_context(
...
+                    &receiver_account_id,
+                )? {
+                    return Ok(crate::commands::PrepopulatedTransaction {
+                        signer_id: signer_account_id.clone(),
+                        receiver_id: ft_contract_account_id.clone(),
+                        actions: vec![],
+                    });
+                }
+
                 let amount_ft = if let crate::types::ft_properties::FungibleTokenTransferAmount::ExactAmount(ft) = &ft_transfer_amount {
```
FroVolod, line 236: Same reason as above

src/commands/tokens/send_near/mod.rs
```diff
@@ -41,7 +41,17 @@ impl From<SendNearCommandContext> for crate::commands::ActionContext {
...
+                    if !crate::common::validate_receiver_account_id(
+                        network_config,
+                        &receiver_account_id,
+                    )? {
+                        return Ok(crate::commands::PrepopulatedTransaction {
+                            signer_id: signer_account_id.clone(),
+                            receiver_id: receiver_account_id.clone(),
+                            actions: vec![],
+                        });
+                    }
```
FroVolod, line 57: Same reason as above

src/commands/tokens/send_nft/mod.rs
```diff
@@ -62,7 +62,17 @@ impl From<SendNftCommandContext> for crate::commands::ActionContext {
...
+                    if !crate::common::validate_receiver_account_id(
+                        network_config,
+                        &receiver_account_id,
+                    )? {
+                        return Ok(crate::commands::PrepopulatedTransaction {
+                            signer_id: signer_account_id.clone(),
+                            receiver_id: nft_contract_account_id.clone(),
+                            actions: vec![],
+                        });
+                    }
```
FroVolod, line 78: Same reason as above

src/commands/tokens/send_ft/amount_ft.rs
Gmin2: Added a verbosity param in `validate_receiver_account_id` to skip validation if in quiet mode to skip validation in https://github.com/near/near-cli-rs/pull/582/commits/fcb7f7345422f147cf59ad1018e05091832b6a97

src/common.rs
```diff
@@ -443,6 +443,90 @@ pub fn find_network_where_account_exist(
...
+) -> color_eyre::eyre::Result<bool> {
+    if let crate::Verbosity::Quiet = verbosity {
+        return Ok(true);
+    }
+    if is_receiver_on_wrong_network(
+        network_config.linkdrop_account_id.as_ref(),
+        receiver_account_id,
+    ) {
+        eprintln!(
+            "\nWarning: <{}> looks like it belongs to a different network than <{}>.",
```
frolvanya: Right now it will throw a warning even for valid implicit accounts, not sure if we want this behaviour

src/common.rs
```diff
@@ -443,6 +443,90 @@ pub fn find_network_where_account_exist(
...
+    linkdrop_account_id: Option<&near_primitives::types::AccountId>,
+    receiver_account_id: &near_primitives::types::AccountId,
+) -> bool {
+    let Some(linkdrop) = linkdrop_account_id else {
+        return false;
+    };
+    let receiver_str = receiver_account_id.as_str();
+    match linkdrop.as_str() {
+        "near" => receiver_str.ends_with(".testnet"),
+        "testnet" => receiver_str.ends_with(".near") && !receiver_str.ends_with(".testnet"),
```
frolvanya: `receiver_str.ends_with(".near")` is the cause of triggering on implicit accounts and accounts like `v1.signer`, `.hot.tg` and etc. I think it's better to check that it just doesn't end with `.testnet`

frolvanya (commented):
Thanks for the contribution!

Found few minor issues, but overall looks good to me. Still probably have to be discussed with a team which option rpc vs string check is better

src/common.rs
```diff
@@ -443,6 +443,90 @@ pub fn find_network_where_account_exist(
...
+    verbosity: crate::Verbosity,
+) -> color_eyre::eyre::Result<bool> {
+    if let crate::Verbosity::Quiet = verbosity {
+        return Ok(true);
+    }
+    if is_receiver_on_wrong_network(
+        network_config.linkdrop_account_id.as_ref(),
+        receiver_account_id,
+    ) {
+        eprintln!(
```
frolvanya: Let's also replace it with `tracing::warn!` to keep consistent UI

src/common.rs
```diff
@@ -443,6 +443,90 @@ pub fn find_network_where_account_exist(
...
+        ));
+
+    match result {
+        Ok(_) => Ok(true),
+        Err(near_jsonrpc_client::errors::JsonRpcError::ServerError(
+            near_jsonrpc_client::errors::JsonRpcServerError::HandlerError(
+                near_jsonrpc_primitives::types::query::RpcQueryError::UnknownAccount { .. },
+            ),
+        )) => {
+            eprintln!(
```
frolvanya: same here

src/common.rs
```diff
@@ -443,6 +443,90 @@ pub fn find_network_where_account_exist(
...
+    let receiver_str = receiver_account_id.as_str();
+    match linkdrop.as_str() {
+        "near" => receiver_str.ends_with(".testnet"),
+        "testnet" => receiver_str.ends_with(".near") && !receiver_str.ends_with(".testnet"),
+        _ => false,
+    }
+}
+
+/// Returns `Ok(true)` if the transaction should proceed, `Ok(false)` if the user cancelled.
+pub fn validate_receiver_account_id(
```
frolvanya, line 470: This method should have `tracing::instrument` procmacro as all other methods that are interacting with RPC to show this info if `--teach-me` flag is provided

src/common.rs
FroVolod: done (5e81a71)

src/common.rs
FroVolod: done (5e81a71)

src/common.rs
FroVolod: done ([bafbb55](https://github.com/near/near-cli-rs/pull/582/commits/bafbb55452a3a02f8492726be69d73db2dab2ee5))

src/common.rs
FroVolod: done (3490ab9)

src/common.rs
FroVolod: done (3490ab9)

src/common.rs
```diff
@@ -443,6 +443,86 @@ pub fn find_network_where_account_exist(
...
+        return ask_if_should_proceed();
+    }
+
+    let result = tokio::runtime::Runtime::new()
+        .unwrap()
+        .block_on(get_account_state(
+            network_config,
+            receiver_account_id,
+            near_primitives::types::BlockReference::latest(),
+        ));
```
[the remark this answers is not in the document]
FroVolod: done (3490ab9)

frolvanya (approved):

vsavchyn-dev (approved):
LGTM, thanks for the PR!

PR #595, opened 2026-05-20
chore: bump social-profile set call gas to 1 PGas

r-near:
## Summary
- Bumps the hardcoded `from_teragas(300)` in `update_social_profile/sign_as.rs` to `from_teragas(1000)` (1 PGas)
- Followup to the 2.11 gas limit increase ([devex#4](https://github.com/near/devex/issues/4)) — this was the last remaining 300 TGas hardcode in the CLI source

## Test plan
- [x] `cargo check` passes
- [ ] Manually verify `near account update-social-profile` still works against social.near


frol:
300 Tgas was already a stretch, I don’t think we need 1000 Tgas

r-near:
> 300 Tgas was already a stretch, I don’t think we need 1000 Tgas

Fair enough, will close this

PR #600, opened 2026-06-03
ci: use npm trusted publishing (OIDC) instead of NPM_TOKEN

r-near:
npm publishing started failing after trusted publishing was enabled for the package on npmjs.com. This switches the `publish-npm` job to OIDC trusted publishing, mirroring the setup in cargo-near:

- Add `id-token: write` permission to the job
- Drop `NODE_AUTH_TOKEN` / `NPM_TOKEN`
- Upgrade npm on the runner (trusted publishing needs npm >= 11.5.1)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -283,6 +283,10 @@ jobs:
       - plan
       - host
     runs-on: "ubuntu-22.04"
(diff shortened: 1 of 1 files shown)


frol:
@r-near There is a cleaner fix: #603

PR #622, opened 2026-07-17
refactor: move generate-keypair under the account entity

r-near:
## What

Moves the offline key generator from the top-level `near generate-keypair` to `near account generate-keypair`, so it sits alongside the other key-related account actions (`get-public-key`, `add-key`, `delete-keys`, …).

## Why

`near-cli-rs` follows a two-level convention: the **top level is entities (nouns)** — `account`, `tokens`, `contract`, … — and the **second level is actions (verbs)**. `generate-keypair` (added in #605) is a verb, so it was out of place at the top level. It belongs under an entity, and `account` is the natural home (from a user's perspective a key ≈ an account).

## Change

- `git mv src/commands/generate_keypair/ → src/commands/account/generate_keypair/` (pure rename, **no logic changes** — it already used `crate::GlobalContext` as its input context, which matches `AccountActions`).
- Removed the top-level module declaration and the `TopLevelCommand::GenerateKeypair` variant.
- Added it as `AccountActions::GenerateKeypair`, placed next to `get-public-key`.

The command is now invoked as `near account generate-keypair …`, and appears under "What do you want to do with an account?" in interactive mode.

## Compatibility

`generate-keypair` only shipped in v0.28.0 (#605) and isn't used anywhere yet, so this is done as a plain move without a deprecation alias.

## Testing

`cargo test --all-features`, `cargo clippy -- -D warnings`, `cargo fmt --check`, and both `--no-default-features` (+`ledger`) checks all pass locally.

## Before / after

It's now an action under `account` (interactive mode):

generate-keypair under the account menu

…and it's gone from the top level, which now lists only entities:

top-level help without generate-keypair

https://github.com/user-attachments/assets/e72a0eef-e794-4441-80cb-d6b81546602e


src/commands/account/mod.rs
```diff
@@ -67,6 +68,11 @@ pub enum AccountActions {
     ))]
     /// Get the public key to your account
     GetPublicKey(self::get_public_key::GetPublicKey),
+    #[strum_discriminants(strum(
```
vsavchyn-dev, line 71: Do we even need `generate_keypair` under `account`? I believe it would make sense to have all the generation only where we need it (aka under `account`->`create-account` -> ... -> `autogenerate-keypair` and `account` -> `add-key` -> ... -> `autogenerate-keypair`).

Additionally, it just doesn't fit under `account` - all actions under it are interactions with account, and `generate_keypair` does not interact w account in any way.

vsavchyn-dev (dismissed):

src/commands/account/mod.rs
r-near: That's fair. I think we need to decide whether offline keygen is something we even want. I'm leaning to just cutting this entirely. 

Because why would someone want to generate a keypair and then _not_ use it?

There was the old `utils generate-keypair` but I guess that wasn't really used much. 

`autogenerate-keypair` already supports ML-DSA separately from this

r-near:
Closing in favor of #623. Per @vsavchyn-dev's review, moving `generate-keypair` under `account` was the wrong call — it doesn't operate on an account. Instead of relocating it, #623 removes the standalone command entirely; ML-DSA-65 key generation stays available through the `add-key` / `create-account` autogenerate flows.

PR #623, opened 2026-07-17
refactor: remove the standalone generate-keypair command

r-near:
## What

Removes the standalone, top-level `near generate-keypair` command that was added in #605.

## Why

`generate-keypair` is an offline, account-less key generator. It doesn't fit the CLI's two-level convention (top level = entities/nouns; `generate-keypair` is a bare verb), and it doesn't fit under any existing entity — as [@vsavchyn-dev noted](https://github.com/near/near-cli-rs/pull/622#discussion_r3606193494), everything under `account` operates *on* an account, which offline keygen never does. (Historically this lived under the old `utils` namespace, which was itself removed in the #115 commands-layout revamp.)

Rather than relocate it, we drop it — the actual key-generation capability, including the new post-quantum **ML-DSA-65** scheme, is already reachable where it's used.

## What is NOT affected

Post-quantum key support is wired independently of this command, via `--signature-scheme` on the autogenerate flows, all of which stay:

- `account add-key … autogenerate-new-keypair`
- `account create-account fund-myself … add-key … autogenerate-new-keypair`
- `account create-account sponsor-by-faucet-service … add-key … autogenerate-new-keypair`

So **creating a (named) account with an ML-DSA-65 key, or adding one to an existing account, works exactly as before.** An externally generated `ml-dsa-65:…` key can still be attached via `add-key … use-manually-provided-public-key`.

The only capability removed is the account-less "generate a keypair and print it / save it to a JSON file" path.

## Details

- Delete `src/commands/generate_keypair/` and its top-level wiring in `src/commands/mod.rs`.
- Keep the shared helpers in `common.rs` (`SignatureScheme`, `GeneratedKeyPair`, `input_signature_scheme`) — the autogenerate flows depend on them.
- Move the ML-DSA-65 / ed25519 key-generation unit tests from the deleted module into `common.rs` (they exercise those helpers).

## Testing

`cargo test --all-features`, `cargo clippy --all-features -- -D warnings`, `cargo fmt --check`, and both `--no-default-features` (+`ledger`) checks pass locally. The 4 moved tests run green at `common::keypair_tests::…`.

Supersedes #622 (which moved the command under `account`); this takes the alternative direction from that review.


src/common.rs
```diff
@@ -3945,3 +3945,87 @@ pub fn parse_borsh_base64_state_init(
     near_primitives::deterministic_account_id::DeterministicAccountStateInit::try_from_slice(bytes)
         .map_err(|e| color_eyre::eyre::eyre!("Failed to borsh-deserialize state init: {e}"))
 }
+
+#[cfg(test)]
+mod keypair_tests {
+    use super::{GeneratedKeyPair, SignatureScheme};
+    use std::str::FromStr;
```
vsavchyn-dev: It looks like it doesn't belong to `common.rs` at all. Also, it's like the only test in `common.rs`.

We can remove it for now or move it where the key gen happens.

r-near: They're the existing #605 tests, just moved so we don't lose the ML-DSA coverage. The code they cover (`GeneratedKeyPair` etc.) lives in this module, so I kept them next to it 🤷 

But yeah, we could probably refactor that code out of here as well, maybe a dedicated keypair module?

vsavchyn-dev: Well, tests should definitely be elsewhere and not in `common.rs` :)
`./tests/unit_tests/generate_keypair.rs` wdyt?

r-near: Done: c02d5480b370e3a472c6fba7d98a3b0ce73f6d9d

Moved it as an integration test against the public API

vsavchyn-dev (approved):

PR #625, opened 2026-07-18
fix: hide NEP-413 private key input

r-near:
## What changed

- use `inquire::Password` when the NEP-413 private key is entered interactively
- keep the existing explicit CLI argument behavior unchanged
- redact secret keys when reconstructing commands for terminal output and CLI history

## Why

The generated `interactive_clap` input path used `inquire::CustomType`, which rendered the private key as it was typed or pasted. The dedicated password prompt hides the input.

A full interactive signing test also showed that the CLI's generated replay command printed the private key after signing and persisted it through the command-history path. Redacting `SecretKey`'s display representation prevents both exposures while leaving parsing and signing unchanged.

This intentionally does not add speculative zeroization or refactor key ownership; the reported clone is required by the generated context scope and no memory-disclosure path was demonstrated.

## Validation

- `cargo fmt --check`
- `cargo check --all-targets`
- `cargo test --all-targets`
- successful NEP-413 signing through the interactive prompt with Terminal Control; input remained hidden and the replay command contained `[REDACTED]`

Fixes #621.

### DIFF
--- a/src/commands/message/sign_nep413/signature_options/sign_with_private_key.rs
+++ b/src/commands/message/sign_nep413/signature_options/sign_with_private_key.rs
@@ -2,12 +2,31 @@
 #[interactive_clap(input_context = super::super::FinalSignNep413Context)]
 #[interactive_clap(output_context = SignPrivateKeyContext)]
 pub struct SignPrivateKey {
+    #[interactive_clap(skip_default_input_arg)]
     /// Enter your private (secret) key:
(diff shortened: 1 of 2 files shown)


r-near:
Closing this. After testing it, I don't think this is the right fix. It treats one private-key prompt differently from the rest of the CLI, and the redaction makes the generated command unusable.

PR #632, opened 2026-07-20
feat: publish Homebrew formula

r-near:
## Summary

- generate a Homebrew formula named `near` with cargo-dist
- publish release formula updates to `near/homebrew-tap`
- document the Homebrew installation command

## Why

This provides an upstream-managed Homebrew package for macOS and Linux. The tap is initialized in https://github.com/near/homebrew-tap/pull/1.

The publisher authenticates with a write-enabled deploy key scoped only to `near/homebrew-tap`; the private key is stored as the `HOMEBREW_TAP_DEPLOY_KEY` Actions secret.

## Validation

- cargo-dist 0.32.0 generated `near.rb` for v0.29.0
- `dist plan` includes the Homebrew artifact and publisher
- `cargo metadata --no-deps --format-version=1`
- `actionlint` (no workflow/expression errors; generated shell snippets retain existing informational ShellCheck findings)
- `git diff --check`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -287,6 +287,52 @@ jobs:
 
           gh release create "${{ needs.plan.outputs.tag }}" --target "$RELEASE_COMMIT" $PRERELEASE_FLAG --title "$ANNOUNCEMENT_TITLE" --notes-file "$RUNNER_TEMP/notes.txt" artifacts/*
 
+  publish-homebrew-formula:
+    needs:
+      - plan
+      - host
+    runs-on: "ubuntu-22.04"
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      PLAN: ${{ needs.plan.outputs.val }}
+      GITHUB_USER: "near-cli-rs release bot"
+      GITHUB_EMAIL: "github-actions[bot]@users.noreply.github.com"
+    if: ${{ !fromJson(needs.plan.outputs.val).announcement_is_prerelease || fromJson(needs.plan.outputs.val).publish_prereleases }}
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          persist-credentials: true
+          repository: "near/homebrew-tap"
+          ssh-key: ${{ secrets.HOMEBREW_TAP_DEPLOY_KEY }}
+      # So we have access to the formula
+      - name: Fetch homebrew formulae
+        uses: actions/download-artifact@v8
+        with:
+          pattern: artifacts-*
+          path: Formula/
+          merge-multiple: true
+      # This is extra complex because you can make your Formula name not match your app name
+      # so we need to find releases with a *.rb file, and publish with that filename.
+      - name: Commit formula files
+        run: |
+          git config --global user.name "${GITHUB_USER}"
+          git config --global user.email "${GITHUB_EMAIL}"
+
+          for release in $(echo "$PLAN" | jq --compact-output '.releases[] | select([.artifacts[] | endswith(".rb")] | any)'); do
+            filename=$(echo "$release" | jq '.artifacts[] | select(endswith(".rb"))' --raw-output)
+            name=$(echo "$filename" | sed "s/\.rb$//")
+            version=$(echo "$release" | jq .app_version --raw-output)
+
+            export PATH="/home/linuxbrew/.linuxbrew/bin:$PATH"
+            brew update
+            # We avoid reformatting user-provided data such as the app description and homepage.
+            brew style --except-cops FormulaAudit/Homepage,FormulaAudit/Desc,FormulaAuditStrict --fix "Formula/${filename}" || true
+
+            git add "Formula/${filename}"
+            git commit -m "${name} ${version}"
+          done
+          git push
+
   publish-npm:
     needs:
       - plan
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -287,6 +287,52 @@ jobs:
 
           gh release create "${{ needs.plan.outputs.tag }}" --target "$RELEASE_COMMIT" $PRERELEASE_FLAG --title "$ANNOUNCEMENT_TITLE" --notes-file "$RUNNER_TEMP/notes.txt" artifacts/*
 
+  publish-homebrew-formula:
```
vsavchyn-dev, line 290: Just curious, can't we use `brew bump-formula-pr`? Or is this tool supposed to be used inside [near/homebrew-tap](https://github.com/near/homebrew-tap) ?

[link to mention of this tool in brew repo](https://github.com/Homebrew/brew/blob/HEAD/docs/How-To-Open-a-Homebrew-Pull-Request.md#submit-a-new-version-of-an-existing-formula) 
[link to manpage of this tool](https://github.com/Homebrew/brew/blob/df30fd34cc7132abfb8dbe3b1d046e3d48a57d00/docs/Manpage.md#bump-formula-pr-options-formula)

.github/workflows/release.yml
```diff
@@ -287,6 +287,52 @@ jobs:
...
+            name=$(echo "$filename" | sed "s/\.rb$//")
+            version=$(echo "$release" | jq .app_version --raw-output)
+
+            export PATH="/home/linuxbrew/.linuxbrew/bin:$PATH"
+            brew update
+            # We avoid reformatting user-provided data such as the app description and homepage.
+            brew style --except-cops FormulaAudit/Homepage,FormulaAudit/Desc,FormulaAuditStrict --fix "Formula/${filename}" || true
+
+            git add "Formula/${filename}"
+            git commit -m "${name} ${version}"
```
vsavchyn-dev, line 332: Just correct me if I'm wrong here:
- We download ALL artifacts from "build artifacts" stage (including .rb file)
- We overwrite previous `Formula/near.rb`
- We commit and push it to the `near/homebrew-tap` repo
- It is somehow pushed to `Homebrew/homebrew-core/n/near.d`? (via `brew bump-formula-pr`?)

.github/workflows/release.yml
```diff
@@ -287,6 +287,52 @@ jobs:
...
+          path: Formula/
+          merge-multiple: true
+      # This is extra complex because you can make your Formula name not match your app name
+      # so we need to find releases with a *.rb file, and publish with that filename.
+      - name: Commit formula files
+        run: |
+          git config --global user.name "${GITHUB_USER}"
+          git config --global user.email "${GITHUB_EMAIL}"
+
+          for release in $(echo "$PLAN" | jq --compact-output '.releases[] | select([.artifacts[] | endswith(".rb")] | any)'); do
```
vsavchyn-dev, line 321: Why do we have a for loop here? Aren't we getting singular `near.rb` artifact to commit?

.github/workflows/release.yml
```diff
@@ -287,6 +287,52 @@ jobs:
...
+        with:
+          persist-credentials: true
+          repository: "near/homebrew-tap"
+          ssh-key: ${{ secrets.HOMEBREW_TAP_DEPLOY_KEY }}
+      # So we have access to the formula
+      - name: Fetch homebrew formulae
+        uses: actions/download-artifact@v8
+        with:
+          pattern: artifacts-*
+          path: Formula/
```
vsavchyn-dev, line 312: Maybe we should put artifacts into something like `./tmp/artifacts`? That mentioned, I think there is no harm in having them there - we are only committing `*.rb` files anyway and it's the same as `publish-npm`.

.github/workflows/release.yml
vsavchyn-dev: I see it's the same as in `publish-npm` action, but still curious on why do we need it :)

README.md
```diff
@@ -12,6 +12,14 @@ Just run `near` and let it guide you through!
 
 Visit [Releases page](https://github.com/near/near-cli-rs/releases/) to see the latest updates.
 
+<details>
+  <summary>Install via Homebrew (macOS and Linux)</summary>
+
+```sh
+brew install near/tap/near
+```
```
vsavchyn-dev, line 20: nit: [brew formulae suggests using](https://formulae.brew.sh/formula/near#default) `brew install near` - I think we should mention that tap is to install binary and suggest using formula as a fallback.

r-near: Yes good catch, I had this before we got published

.github/workflows/release.yml
r-near: Probably. FYI this whole job is auto-generated by cargo-dist: https://github.com/axodotdev/cargo-dist/blob/main/cargo-dist/templates/ci/github/partials/publish_homebrew.yml.j2

.github/workflows/release.yml
r-near: Tbh it's just cargo-dist. I didn't spend time optimizing this at all

r-near:
Closing: `near` is now in homebrew-core (https://formulae.brew.sh/formula/near) and gets autobumped on new releases, so we don't need to maintain a separate tap. README update in #642.

PR #642, opened 2026-08-17
docs: document Homebrew installation

r-near:
## Summary

- add `brew install near` to the README install options

## Why

`near` is published in homebrew-core (https://formulae.brew.sh/formula/near) with bottles for macOS and Linux, and Homebrew autobumps it on new releases. That makes the `near/homebrew-tap` publisher from #632 unnecessary, so this replaces that PR with just the docs change.


vsavchyn-dev (approved):
doc change lgtm :)

PR #644, opened 2026-08-19
fix: paginate view_state queries so large contract state can be read

r-near:
`near contract view-storage` fails with `TOO_LARGE_CONTRACT_STATE` on any account whose state exceeds the ~50 KB `view_state` cap that public RPCs enforce (`trie_viewer_state_size_limit`). The same one-shot query is used to list active staking pools for `near account view-account-summary` when no `fastnear_url` is configured, and `poolv1.near` is well past that cap, so that fallback fails too.

nearcore 2.13 added pagination to `view_state` (`limit` / `after_key_base64` → `last_key`), and paginated requests are not subject to the size cap. This adds `common::fetch_contract_state`, which always requests pages and follows the cursor until the listing is complete, and uses it from both call sites. Pages after the first are read at the block hash the first page resolved to, so `now` still yields a consistent snapshot. Older RPC nodes ignore the paging fields and return the whole (capped) state without a cursor, so behaviour there is unchanged.

Since `include_proof` was already hard-coded to `false`, the always-empty `Contract state (proof)` line is dropped from the `view-storage` output.

Verified against mainnet: `view-storage farm.berryclub.ek.near all` (7 pages, 2,799 entries, matches a raw paginated RPC walk) and `usn` (3,391 entries), both of which fail on the current release; `view-account-summary` without `fastnear_url` now lists all 556 `poolv1.near` pools.

### DIFF
--- a/src/commands/contract/view_storage/output_format/as_json.rs
+++ b/src/commands/contract/view_storage/output_format/as_json.rs
@@ -20,17 +20,9 @@ impl AsJsonContext {
             let prefix = previous_context.prefix;
 
             move |network_config, block_reference| {
-                let query_view_method_response =
+                let values =
                     super::get_contract_state(&contract_account_id, prefix.clone(), network_config, block_reference.clone())?;
-
-                if let near_jsonrpc_primitives::types::query::QueryResponseKind::ViewState(result) =
-                    query_view_method_response.kind
-                {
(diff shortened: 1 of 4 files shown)


r-near:
Superseded by #587, which migrates these call sites to near-kit — the same fix (always request `view_state` pages and follow the cursor) is now on that branch as 9b50103, so this standalone version would only conflict with it.

PR #645, opened 2026-08-19
refactor: `account import-account` command

vsavchyn-dev:
## What is this PR
This PR reworks `account import-account` so the whole flow can be expressed on the command line, with the interactive as a fallback rather than a requirement.

- Account ID is actually present as an argument after inputs for one of the `import-account` commands and before `network-config` selection, instead of being prompted for mid-flow during `network-config` selection.
- `network-config` step now owns the rest of the flow: the account check and a new save-mode subcommands.
- Account/key verification is opt-in via `--check-account-id`, instead of an unconditional loop that re-asked for the account ID whenever verification failed.
- Explicit save modes: `save-to-keychain`, `save-to-legacy-keychain`, `print-to-terminal`. Previously the keychain choice was an `inquire::Select` with no `ToCli` representation at all, so an import could never be scripted end to end. `print-to-terminal` is new.

## Before / after
before - two prompts (account ID without verification loop, keychain choice) that no flag could satisfy
```
near account import-account using-private-key ed25519:... network-config testnet
```

after - fully scriptable
```
near account import-account using-private-key ed25519:... alice.testnet \
  network-config testnet --check-account-id save-to-keychain
```

## Interactive vs scripted

`--check-account-id` is a presence flag, and `NetworkForImportAccount` uses hand-written `FromCli`, so it can tell the two situations apart:

- network name and save mode both supplied on the command line? -> treat as scripted: the question is never asked, and the check runs only if `--check-account-id` was passed.
- anything missing? -> ask "Would you like to check if account <X> exists?", and the answer is reflected in the `Here is your console command...` line at the end.

### DIFF
--- a/src/commands/account/import_account/mod.rs
+++ b/src/commands/account/import_account/mod.rs
@@ -1,5 +1,4 @@
-#![allow(clippy::enum_variant_names, clippy::large_enum_variant)]
-use std::{str::FromStr, vec};
+use std::str::FromStr;
 
 use color_eyre::eyre::Context;
 use inquire::{CustomType, Select};
@@ -7,27 +6,52 @@ use strum::{EnumDiscriminants, EnumIter, EnumMessage};
 
 use near_primitives::account::id::AccountType;
 
+pub mod key_store_prop;
+mod network;
 mod using_private_key;
 mod using_seed_phrase;
 mod using_web_wallet;
 
 #[derive(Debug, Clone, interactive_clap::InteractiveClap)]
-#[interactive_clap(context = crate::GlobalContext)]
+#[interactive_clap(input_context = crate::GlobalContext)]
+#[interactive_clap(output_context = ImportAccountCommandContext)]
 pub struct ImportAccountCommand {
+    /// Enter account ID to import:
+    account_id: crate::types::account_id::AccountId,
+
     #[interactive_clap(subcommand)]
+    /// How would you like to import the account?
     import_account_actions: ImportAccountActions,
 }
 
+#[derive(Debug, Clone)]
+pub struct ImportAccountCommandContext {
+    global_context: crate::GlobalContext,
+    account_id: near_primitives::types::AccountId,
+}
+
+impl ImportAccountCommandContext {
+    pub fn from_previous_context(
+        previous_context: crate::GlobalContext,
+        scope: &<ImportAccountCommand as interactive_clap::ToInteractiveClapContextScope>::InteractiveClapContextScope,
+    ) -> color_eyre::eyre::Result<Self> {
+        Ok(Self {
+            global_context: previous_context,
+            account_id: scope.account_id.clone().into(),
+        })
+    }
+}
+
 #[derive(Debug, EnumDiscriminants, Clone, interactive_clap::InteractiveClap)]
-#[interactive_clap(context = crate::GlobalContext)]
+#[interactive_clap(context = ImportAccountCommandContext)]
 #[strum_discriminants(derive(EnumMessage, EnumIter))]
 /// How would you like to import the account?
 pub enum ImportAccountActions {
-    #[strum_discriminants(strum(
-        message = "using-web-wallet          - Import existing account using NEAR Wallet (a.k.a. \"sign in\")"
-    ))]
-    /// Import existing account using NEAR Wallet (a.k.a. "sign in")
-    UsingWebWallet(self::using_web_wallet::LoginFromWebWallet),
+    // #[strum_discriminants(strum(
+    //     message = "using-web-wallet          - Import existing account using NEAR Wallet (a.k.a. \"sign in\")"
+    // ))]
+    // /// Import existing account using NEAR Wallet (a.k.a. "sign in")
+    // UsingWebWallet(self::using_web_wallet::LoginFromWebWallet),
     #[strum_discriminants(strum(
         message = "using-seed-phrase         - Import existing account using a seed phrase"
     ))]
@@ -39,6 +63,50 @@ pub enum ImportAccountActions {
     /// Import existing account using a private key
     UsingPrivateKey(self::using_private_key::LoginFromPrivateKey),
 }
+
+// #[derive(Debug, Clone, interactive_clap::InteractiveClap)]
+// #[interactive_clap(input_context = ImportAccountKeyContext)]
+// #[interactive_clap(output_context = ImportAccountDetailsContext)]
+// pub struct ImportAccountDetails {
+//     #[interactive_clap(named_arg)]
+//     /// Select network:
+//     network: network::NetworkForImportAccountArgs,
+// }
+//
+// #[derive(Debug, Clone)]
+// pub struct ImportAccountDetailsContext {
+//     global_context: crate::GlobalContext,
+//     key_pair_properties_buf: key_store_prop::KeyPairProperty,
+//     public_key_str: String,
+//     account_id: near_primitives::types::AccountId,
+// }
+//
+// impl ImportAccountDetailsContext {
+//     pub fn from_previous_context(
+//         previous_context: ImportAccountKeyContext,
+//         scope: &<ImportAccountDetails as interactive_clap::ToInteractiveClapContextScope>::InteractiveClapContextScope,
+//     ) -> color_eyre::eyre::Result<Self> {
+//         Ok(Self {
+//             global_context: previous_context.global_context,
+//             key_pair_properties_buf: previous_context.key_pair_properties_buf,
+//             public_key_str: previous_context.public_key_str,
+//             account_id: scope.account_id.clone().into(),
+//         })
+//     }
+// }
+
+// impl From<ImportAccountKeyContext> for crate::network::NetworkContext {
+//     fn from(item: ImportAccountKeyContext) -> Self {
+//         let on_after_getting_network_callback: crate::network::OnAfterGettingNetworkCallback =
+//             std::sync::Arc::new({ move |network_config| Ok(()) });
+//
+//         Self {
+//             config: item.global_context.config,
+//             interacting_with_account_ids: vec![item.account_id],
+//             on_after_getting_network_callback,
+//         }
+//     }
+// }
 
 pub fn login(
     network_config: crate::config::NetworkConfig,
--- a/src/commands/account/import_account/mod.rs
+++ b/src/commands/account/import_account/mod.rs
@@ -72,6 +72,31 @@ pub enum ImportAccountActions {
     UsingPrivateKey(self::using_private_key::LoginFromPrivateKey),
 }
 
+fn check_implicit_account_id(
+    account_id: &near_primitives::types::AccountId,
+    public_key: &near_crypto::PublicKey,
+) -> color_eyre::eyre::Result<()> {
+    if near_primitives::account::id::AccountType::NearImplicitAccount
+        != account_id.get_account_type()
+    {
+        return Ok(());
+    }
+
+    let expected_public_key = near_crypto::PublicKey::from_near_implicit_account(account_id)
+        .map_err(|err| {
+            color_eyre::Report::msg(format!(
+                "Failed to derive the public key of implicit <{account_id}>: {err}"
+            ))
+        })?;
+
+    if *public_key != expected_public_key {
+        return Err(color_eyre::eyre::eyre!(
+            "Provided public key cannot sign for implicit account <{account_id}>:\n   expected public key: {expected_public_key}\n   provided public key: {public_key}"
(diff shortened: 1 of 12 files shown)


src/commands/account/import_account/using_web_wallet/mod.rs
```diff
@@ -35,31 +47,14 @@ impl LoginFromWebWalletContext {
...
-                        url.as_str()
-                    );
-                    super::login(
-                        network_config.clone(),
-                        config.credentials_home_dir.clone(),
-                        &key_pair_properties_buf,
-                        &key_pair_properties.public_key_str,
-                        &error_message,
-                    )
+                    Ok(key_store_property)
```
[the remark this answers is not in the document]
vsavchyn-dev: Welp, we are doing the sign-in with wallet, so waiting for it might take some time and I don't think it's worth blocking cli on this.

src/commands/account/import_account/mod.rs
```diff
@@ -40,121 +72,37 @@ pub enum ImportAccountActions {
...
-pub fn login(
-    network_config: crate::config::NetworkConfig,
-    credentials_home_dir: std::path::PathBuf,
-    key_pair_properties_buf: &str,
-    public_key_str: &str,
-    error_message: &str,
+pub fn check_account_id(
+    global_context: &crate::GlobalContext,
+    chosen_network_config: &crate::config::NetworkConfig,
+    account_id: &near_primitives::types::AccountId,
```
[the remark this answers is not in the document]
vsavchyn-dev: Good catch - now we check the implicit account id in "using-private-key" and "using-seed-phrase" https://github.com/near/near-cli-rs/pull/645/commits/f3dbf357332f4f2ff2eab3fad34288d935917750

For using-web-wallet - it's a sign in... Not sure what to do there

src/commands/account/import_account/network/save_keypair_to_keychain/mod.rs
```diff
@@ -0,0 +1,27 @@
...
+        println!(
+            "{}",
+            crate::common::save_access_key_to_keychain_or_save_to_legacy_keychain(
+                previous_context.chosen_network_config,
+                previous_context.config.credentials_home_dir,
+                &serde_json::to_string(&previous_context.key_store_property)?,
+                &previous_context.key_store_property.to_public_key_str(),
+                previous_context.account_id.as_str(),
+            )?
+        );
```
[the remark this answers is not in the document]
vsavchyn-dev: Now we register account ids as signer (https://github.com/near/near-cli-rs/pull/645/commits/f3dbf357332f4f2ff2eab3fad34288d935917750)

src/commands/account/import_account/network/print_keypair_to_terminal/mod.rs
```diff
@@ -0,0 +1,52 @@
...
+        display_string.push_str(&format!(
+            "\n{}",
+            crate::common::indent_payload(&display_info)
+        ));
+
+        tracing::info!(
+             parent: &tracing::Span::none(),
+             "{}",
+             display_string
+        );
```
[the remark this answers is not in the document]
vsavchyn-dev: Not sure it's a valid concern - I would people to use `print-to-terminal` in non-quiet mode. If you want to print things - you want them printed out normally, no?

src/commands/account/import_account/mod.rs
```diff
@@ -40,121 +72,62 @@ pub enum ImportAccountActions {
...
-            crate::common::save_access_key_to_keychain_or_save_to_legacy_keychain(
-                network_config,
-                credentials_home_dir,
-                key_pair_properties_buf,
-                public_key_str,
-                account_id.as_ref(),
-            )?;
-        eprintln!("{storage_message}");
+    // Implicit account always exists - we import it without prompting to check if it exists.
+    if let AccountType::NearImplicitAccount = account_id.get_account_type() {
```
vsavchyn-dev:
```suggestion
    if account_id.get_account_type().is_implicit() {
```
I am not sure what check should be applied here, but considering that we are checking if account is implicit, we probably should use `AccountType::is_implicit()` here.

src/commands/account/import_account/mod.rs
```diff
@@ -40,121 +72,62 @@ pub enum ImportAccountActions {
...
-                break account_id_from_cli;
-            }
-        };
+fn check_implicit_account_id(
+    account_id: &near_primitives::types::AccountId,
+    public_key: &near_crypto::PublicKey,
+) -> color_eyre::eyre::Result<()> {
+    if near_primitives::account::id::AccountType::NearImplicitAccount
+        != account_id.get_account_type()
+    {
```
vsavchyn-dev: This check should probably cover `AccountType::EthImplicitAccount` as well?

frol (dismissed):
@vsavchyn-dev It was designed to be an interactive flow since in most of the cases users who need to import an account would do it interactively anyway. For automations, it would make more sense to supply the key during the transaction signing directly instead of adding the import account step.

> Account ID moves to the front of the command, as an argument of import-account, instead of being prompted for mid-flow after the key was already obtained.

As a user of an implicit account id I would find it daunting to figure out my account id when I only recorded my seed phrase. At least move it after the key details.

> Change network-config -> network, and that step now owns the rest of the flow: the account check and a new save-mode subcommands.

"testnet-fastnear" is not a network, so changing "network-config" to "network" is not the best idea IMO. Also, it will break all the existing scripted commands, right?

> so an import could never be scripted end to end. print-to-terminal is new.

What is the kind of user do you optimize it for? I don't see the use-case for scripting the import account flow.

vsavchyn-dev:
> As a user of an implicit account id I would find it daunting to figure out my account id when I only recorded my seed phrase. At least move it after the key details.

I was doubting wether i should put account id input after or before selecting import method... My initial thought was that people would enter account id that they want to import first. But you are right - it's better to have account-id after choosing a way to import account and provide option to enter implicit account-id.

>  "testnet-fastnear" is not a network, so changing "network-config" to "network" is not the best idea IMO. Also, it will break all the existing scripted commands, right?

Technically, I am not sure if there are "existing scripted commands" considering that it was originally designed to be interactive, so by renaming it I don't think anything will break (and I thought it will be released as a breaking change anyway). However, renaming "network-config" wasn't a right call for sure - I double-checked and I don't see usage of `network` at all. I will revert the naming.

> What is the kind of user do you optimize it for? I don't see the use-case for scripting the import account flow.

From what I understand the idea is to import account and then sign some of the migrations on localnet/testnet. 
For instance:
```
near account import-account <account-id> using-private-key <pk> network testnet save-to-keychain
near contract call-function as-transaction <account-id> <func_name> json-args '{"abc": "cba"}' prepaid-gas '100.0 Tgas' attached-deposit '1 NEAR' sign-as <account-id> network-config testnet sign-with-keychain send
```
However, I totally forgot that there is `transaction construct-transaction` and `sign-with-plaintext-private-key`...

Still, I think for usecase `import-account` and then `construct-transaction ... sign-with-keychain` it might make sense to have it scriptable. Not as efficient as providing private key during signing, but you might want to store account in keychain after signing transaction, right?

cc @frol @SimonRastikian (on info about the usecase)

frol:
> but you might want to store account in keychain after signing transaction, right?

Why would anyone want to spam their global system keychain with scripted keys?

As a rule of thumb, unless you have a good usecase in mind for the actual users of the tool, don’t bring features that only contribute complexity. Sure you can build anything, but ask yourself why it is actually needed before jumping on it.

vsavchyn-dev:
> Why would anyone want to spam their global system keychain with scripted keys?

Fair, nobody :) My comment was misleading - I didn't have a concrete example back then.

First one would be setting up a localnet for MPC - a single `just setup-localnet` that spins up localnet in docker, creates the dev accounts, and imports them. Then, you can run `near ... sign-with-keychain` interactively. Those aren't scripted keys spamming a global store - they're the same keys the developer would otherwise sit and enter by hand.

Next example is batch import - pulling N accounts in from a list of seed phrases, or JS CLI credentials. Destination is still the developer's own keychain, it's just that the interactive flow doesn't scale to 20 accounts.

> As a rule of thumb, unless you have a good usecase in mind for the actual users of the tool, don’t bring features that only contribute complexity. Sure you can build anything, but ask yourself why it is actually needed before jumping on it.

Well, I agree that I didn't have concrete examples back then, so it was hard to justify this feature. However, I am more than sure this feature is useful considering examples I provided above.

Also, I am not sure what do you mean by `feature that contributes only complexity` - most of the code is a boilerplate that `ToCli` would've generated anyways and we can see similar implementation in `network_for_transaction/mod.rs`. Only few things were "new" as I see: `is_scripted` check and prompt to check account_id. Maybe I am missing something here?

src/commands/account/import_account/mod.rs
frol: Given that the access key that matches the implicit account id can be removed after initialization, we may want to always verify that the access key exists on chain and in case if it does not exist, do NOT raise the error ONLY if it is a NEAR implicit account id matching the ed25519 key - this is the only implicit account id that would work with the native NEAR access key right away after getting initialized (at least before 0u implicit accounts with ml-dsa keys are live).

src/commands/account/import_account/mod.rs
frol: It depends on what it is used for. See my other comment.

I don't think near.cli.rs supports EVM access keys in the first place, so we should not cover EthImplicitAccount here as importing 0xdeadbeef with some ed25519 key won't be a right match.

src/commands/account/import_account/network/mod.rs
```diff
@@ -0,0 +1,221 @@
...
+
+impl NetworkForImportAccount {
+    fn input_check_account_id(
+        context: &NetworkForImportAccountContext,
+    ) -> color_eyre::eyre::Result<bool> {
+        use inquire::Select;
+        use near_primitives::account::id::AccountType;
+
+        // Implicit accounts always exist - nothing to verify
+        if let AccountType::NearImplicitAccount = context.account_id.get_account_type() {
```
frol: It is not always true. Also, keep in mind that implicit account can also have other keys besides the default one, and the default one can also be deleted.

frol (commented):
The updated structure makes sense to me. Yet, see my comments regarding the implicit accounts.

frol:
@vsavchyn-dev The presented usecases are valid. Yet, they are not dominating ones, so don't let them to break the regular flow. That said, the latest changes (which are not fully reflected in the PR description) look good to me and they don't break the regular flow.

src/commands/account/import_account/mod.rs
vsavchyn-dev: Okay so from what I understand:
- Account doesn't exist and pubkey from implicit acc == pubkey passed -> it's fine as you can control the acc still
- Account exists but pubkey doesn't exist on implicit acc -> throw error as you cannot control account

Correct me if I`m wrong please
Here is the fix: [a597a7e](https://github.com/near/near-cli-rs/pull/645/commits/a597a7ee18b897afe25e33db8e4f5089596996e2)

src/commands/account/import_account/mod.rs
vsavchyn-dev: I ended up removing this `check_implicit_account_id` considering that `check_account_id` takes care of checking implicit account. With your previous comment in mind added warning about implicit account and pubkey missmatch if person doesn't check the account_id: [a597a7e](https://github.com/near/near-cli-rs/pull/645/commits/a597a7ee18b897afe25e33db8e4f5089596996e2)

src/commands/account/import_account/network/mod.rs
vsavchyn-dev: Fixed in [a597a7e](https://github.com/near/near-cli-rs/pull/645/commits/a597a7ee18b897afe25e33db8e4f5089596996e2)

src/commands/account/import_account/mod.rs
frol: That is correct, and the fix looks good to me

frol (approved):

src/commands/account/import_account/network/mod.rs
```diff
@@ -0,0 +1,220 @@
...
+                &network_config,
+                &previous_context.account_id,
+                previous_context.key_store_property.public_key(),
+            )?;
+        } else {
+            super::warn_on_implicit_account_id_missmatch(
+                &previous_context.account_id,
+                previous_context.key_store_property.public_key(),
+            );
+        }
```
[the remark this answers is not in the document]
vsavchyn-dev: This is reasonable considering implicit account might have different public key from the one used to create it. it shouldn't be a hard error.

PR #650, opened 2026-08-29
fix: support ML-DSA key handles offline

r-near:
## Summary

- accept full public keys or canonical key handles for offline keychain signing
- include every supported key type in the offline legacy-keychain picker

## Verification

- `cargo fmt --all -- --check`
- `cargo test --lib`
- `cargo test --test generate_keypair`
- `cargo clippy -- -D warnings`
- signed offline transactions using an ML-DSA handle, the expanded public key, and the interactive ML-DSA picker

Depends on #649.

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


vsavchyn-dev (commented):
I believe that people MUST enter full public key - in this way, people ensure they have a full access to the ml-dsa key and can somehow recover it later. Still, being able to input pubkey hash is nice for DevEx, so I am not sure what would be better here....

r-near:
I agree it's probably better UX to keep the handle. I imagine most people will just check nearblocks or an indexer (and they don't store full keys) 🤷

vsavchyn-dev (approved):
Yeah, you're right. It's better to just have ability for people to write down the hash that is stored inside the keychain if needed

PR #658, opened 2026-09-02
feat: make global hash deployment retry-safe

r-near:
`contract deploy-as-global use-file <wasm> as-global-hash` re-deploys even when that exact code hash is already on chain. The runtime accepts the duplicate and burns the global-contract storage fee again, so re-running the command after a timeout or from CI costs NEAR for nothing.

Now the `as-global-hash` path prints the Wasm code hash and queries `view_global_contract_code` for it at final state before building the transaction. If the code is already there, the CLI prints "already exists on <network> network. No transaction needed." and exits without signing anything. Any other query failure aborts the command. `as-global-account-id` and `transaction construct-transaction` are unchanged.

In offline mode the check is skipped with a warning and the transaction is built as before, the same way `state-init` and account validation handle `--offline`.

Known gap: two publishers racing can both see "absent" and both pay, and a deploy that is not yet final is also still seen as absent.

To verify: `cargo test --test deploy_global` deploys the same wasm twice against a local sandbox and checks that the second run prints the skip message and sends nothing.


src/commands/contract/deploy_global/mod.rs
```diff
@@ -71,17 +73,23 @@ impl DeployGlobalModeContext {
         previous_context: ContractFileContext,
         scope: &<DeployGlobalMode as interactive_clap::ToInteractiveClapContextScope>::InteractiveClapContextScope,
     ) -> color_eyre::eyre::Result<Self> {
+        let mode = match scope {
+            DeployGlobalModeDiscriminants::AsGlobalHash => {
+                if previous_context.global_context.offline {
+                    return Err(color_eyre::eyre::eyre!(
+                        "`as-global-hash` checks whether the code hash is already deployed, which requires network access. Use `transaction construct-transaction ... add-action deploy-global-contract ... as-global-hash` to build the transaction offline without that check."
```
vsavchyn-dev: I really don't like that we throw error when user is in offline mode. It is a breaking change, and I would rather have warning and "reject" only when user agrees to this behaviour.

vsavchyn-dev: Also, I personally see `construct-transaction` as a wrapper around `tokens ... send-near`, `contract deploy`, or similar flows that just bundles them together into a nice single transaction. With that, I think we shouldn't make `contract deploy-as-global ... as-global-hash` that much different from `transaction construct-transaction ... add-action deploy-global-contract ... as-global-hash`.

src/commands/contract/deploy_global/mod.rs
```diff
@@ -139,24 +147,76 @@ impl DeployGlobalResultContext {
...
+                        near_jsonrpc_client::methods::query::RpcQueryRequest {
+                            block_reference: near_primitives::types::Finality::Final.into(),
+                            request: near_primitives::views::QueryRequest::ViewGlobalContractCode {
+                                code_hash,
+                            },
+                        },
+                    );
+                    if global_contract_exists(code_hash, query_result)? {
+                        eprintln!("Global contract <{code_hash}> already exists; skipping deployment.");
+                        actions.clear();
```
vsavchyn-dev: I think we shouldn't return empty actions array, and just provide the user with warning inside `on_after_getting_network_callback`. See my comment above.

vsavchyn-dev (changes requested):
The idea is nice, but while I do see need for this behaviour to prevent users from spending NEARs on already-deployed global-contract-by-hash, I don't think we should throw error and reject the tx, but rather warn a user or make this "check & reject" behaviour opt-in.

src/commands/contract/deploy_global/mod.rs
r-near: Yeah good point, it does break `sign-later`... 

I'll make offline skip the check with a warning

src/commands/contract/deploy_global/mod.rs
r-near: state-init already cancels with empty actions when the account exists, so we do have some precedent... 

But also, what's the scenario for sending a duplicate tx? We could add an opt-in flag but idk if that'd be useful?

src/commands/contract/deploy_global/mod.rs
```diff
@@ -139,24 +141,89 @@ impl DeployGlobalResultContext {
...
+                        );
+                    } else {
+                        let query_result = network_config.json_rpc_client().blocking_call(
+                            near_jsonrpc_client::methods::query::RpcQueryRequest {
+                                block_reference: near_primitives::types::Finality::Final.into(),
+                                request: near_primitives::views::QueryRequest::ViewGlobalContractCode {
+                                    code_hash,
+                                },
+                            },
+                        );
```
vsavchyn-dev: Can't we move querying to `global_contract_exists`?

src/commands/contract/deploy_global/mod.rs
```diff
@@ -139,24 +141,89 @@ impl DeployGlobalResultContext {
     }
 }
 
+/// Whether global contract code hashing to `expected_hash` is already deployed.
+fn global_contract_exists(
```
vsavchyn-dev:
```suggestion
fn global_contract_by_hash_exists(
```
More of a nit, as I would expect `global_contract_exists` to check if both global contract by account-id  and hash exist (based on `GlobalContractDeployMode`)

vsavchyn-dev (changes requested):
Few nits

src/commands/contract/deploy_global/mod.rs
vsavchyn-dev: What if someone wants to just waste their NEARs? :P 
But yeah you're right, there is no point in making this an opt-in behaviour if users will just loose their money. I was aiming to preserve legacy behaviour, so it wouldn't be a breaking change (again).

src/commands/contract/deploy_global/mod.rs
r-near: Ok cleaned it up here: [`471a7c6` (this PR)](https://github.com/near/near-cli-rs/pull/658/commits/471a7c6488d8a6b90d931a3e1d70dc3231a897c5)

Also moved the query into the helper

r-near: Also added a sandbox integration test instead of a unit test

vsavchyn-dev (approved):
LGTM

PR #671, opened 2026-09-14
chore: switch from fast_clap fork back to upstream clap 4.6.7

r-near:
## Summary

Switches the `clap` dependency from the temporary `fast_clap` fork back to upstream `clap` 4.6.7. This is step 5 ("Switch back to upstream clap") of https://github.com/near/devex/issues/40.

#557 moved us onto `fast_clap` to get lazy/deferred subcommand initialization, which fixed the 5–7 s startup caused by eagerly building the deeply nested derived subcommand tree. That work has now landed upstream:

- clap-rs/clap#6521 — `feat(derive): Defer enum subcommand initialization` (adds `#[command(defer = <bool>)]` on `Parser`/`Subcommand` derives; opt-in in v4, on by default under `unstable-v5`)
- Released in [clap v4.6.7](https://github.com/clap-rs/clap/releases/tag/v4.6.7)

## Changes

- `Cargo.toml`: `fast_clap = { version = "4.5.57", ... }` → `clap = { version = "4.6.7", features = ["derive", "unstable-v5"] }`
- `Cargo.lock`: `fast_clap` / `fast_clap_derive` removed, `clap` / `clap_derive` 4.6.7 added (`cargo tree -i fast_clap` now resolves to nothing)

No source changes were needed:

- The `#[arg(...)]` / `#[command(...)]` attribute modernization from #557 is upstream syntax.
- `interactive-clap` / `interactive-clap-derive` 0.3.2 do not depend on `clap` themselves (`cargo tree -i clap` shows only `near-cli-rs`), and the derives they emit pick up the `unstable-v5` deferral default without an explicit `defer = true`.

## Startup timings

Release builds, macOS, median of 7 runs:

| Command | `main` (fast_clap 4.5.57) | this PR (clap 4.6.7 + `unstable-v5`) |
|---|---|---|
| `near --version` | 8.2 ms | 5.9 ms |
| `near --help` | 5.2 ms | 4.7 ms |
| `near transaction construct-transaction --help` | 4.8 ms | 4.6 ms |
| `construct-transaction … add-action` `--help`, 10 actions deep | 4.9 ms | 4.9 ms |

The 10-deep `add-action transfer 1NEAR` chain (the new limit from #667) parses and prints help correctly at every depth tested (3, 5, 8, 10).

For reference, I also built clap 4.6.7 **without** `unstable-v5` (i.e. no deferral). With the 10-level action chain from #667, `near --version` never finishes: RSS passes 85 GB within 15 s and the process is SIGKILLed after ~115 s on a 128 GB machine. So the `unstable-v5` feature is load-bearing and must stay on until deferral becomes the v5 default.

## Checks

- `cargo build --release` ✅
- `cargo fmt --check` ✅
- `cargo clippy --all-targets -- -D warnings` ✅
- `cargo test` ✅ (52 unit + 20 integration passed, 9 ignored network tests, 0 failed)

Refs: #557, #667, clap-rs/clap#6521, https://github.com/near/devex/issues/40


vsavchyn-dev (approved):
So, on my M4 Pro the build time for main and this PR is the same.

Additionally, startup is the same across main branch and this pr, so I think this pr should be gtg.

cc @frol as we move away from "fast_clap"

